PR #46, opened 2022-06-21
feat: access service

hugomrdias:


Gozala:
@hugomrdias I think it would make more sense to target this `w3` branch

access/packages/api/src/ucanto/capabilities.js
```diff
@@ -0,0 +1,73 @@
...
+    canDelegateURI(child.caveats.as, parent.caveats.as) &&
+    canDelegateURI(child.with, parent.with),
+})
+
+/**
+ * `identity/identify` can be derived from any of the `store/*`
+ * capability that has matichng `with`. This allows store service
+ * to identify account based on any user request.
+ */
+export const identityIdentify = capability({
```
Gozala, line 69: As discussed before this really should be something along the lines of 

https://github.com/web3-storage/ucanto/blob/c73a31ac122af1319183bf3be6447d70e01d6c55/w3/store/src/identity/capability.js#L72-L84

Otherwise "store/*" service provider will not be able to invoke this capability. Alternatively we could delegate "identity/identify" capability from this service to "store/*" service but that is IMO not a better option because:

1. Instead of tracking dependencies in source code and with packages we'd end up doing it through brittle configurations.
2. Connection between "store/*" request and "identity/identify" that it caused will be lost.
3. Delegating unconditional "identity/identify" capability to "store/*" service provider would imply that compromise in later could allow attacker to obtain identity info about all of our accounts. Deriving capability instead would not have that problem as identity could only be resolved for upload target DIDs.

I agree that it is unfortunate that this implies that `identity/identify` has to depend on defined capabilities for `store/*`, but then again this dependency would exist even if we leave it out of code because "identity/*" service needs to know about:

1. Which capabilities it can be derived from.
2. How capabilities it can be derived from can be compared & validated.

Which is precisely what capability definitions do. I think we just need to factor out all the related capabilities into separate module and import them from `store/*` service provider, client and `identity` service provider and client.

access/packages/api/src/ucanto/client-codec.js
```diff
@@ -0,0 +1,24 @@
...
+import { UTF8 } from '@ucanto/transport'
+
+/** @type {import('./types.js').ClientCodec} */
+export const clientCodec = {
+  async encode(invocations, options) {
+    /** @type {Record<string, string>} */
+    const headers = {}
+    const chain = await Delegation.delegate(invocations[0])
+
+    // TODO iterate over proofs and send them too
```
Gozala, line 12: Maybe this route should be omitted until this is finished.

access/packages/api/src/ucanto/service.js
```diff
@@ -0,0 +1,77 @@
...
+            //   Server.UCAN.format(delegation.data)
+            // )
+
+            return {
+              delegation: UCAN.format(delegation.data),
+            }
+          }
+
+          await sendEmail({
+            to: capability.caveats.as.replace('mailto:', ''),
```
Gozala, line 44:
```suggestion
            to: capability.caveats.as.slice('mailto:'.length),
```

Gozala (changes requested):
Provided feedback inline.

access/packages/api/src/ucanto/capabilities.js
hugomrdias: i really dont like this, we should figure out a better option. What about other services that need to call identity/identify but dont deal with store/* ?

Gozala: The thing is services needs to know about each other when they interact in order, there’s no way around this. Now you can thread this knowledge either:

1. Delegate capabilities for interaction (from provider to caller)
2. Define capabilities in provider so caller can just relay user requests.
3. Require that user provide additional capabilities so that service can relay those.

In the end you need to encode that knowledge in one of the three interacting actors and regardless of the choice there are tradeoffs & added complexity.

That said maybe rethinking accounts as we discussed is going to help a bit, specifically linking email increasing account capacity would remove dependency here.
Yet it will create dependency elsewhere vitrification & capacity increase is still going to be distinct capabilities that will either:

1. Be derived from one another
2. Be unlinked from one another

I would still argue that deriving “account/activate” from “identity
/verify” is account services business logic and therefor it makes sense to encode that in its code. It also means that in the future this logic could change without necessarily coordinating with other actors.

If you were to go other way round and just delegate “account/activate” to verifier changing business logic of account service would become more difficult & would require coordinated changes in both services

PR #62, opened 2022-06-25
W3 signer token

ice-breaker-tg:
Extends options for the signer service and the SIGV4 to allow passing the security token generated when a temp IAM is used instead of a user based one.

### DIFF
--- a/w3/sigv4/src/index.js
+++ b/w3/sigv4/src/index.js
@@ -6,7 +6,7 @@ export class SigV4 {
   /**
    * @param {import('./types').SigV4Options} options
    */
-  constructor({ accessKeyId, secretAccessKey, region, cache, securityToken }) {
+  constructor({ accessKeyId, secretAccessKey, region, cache, sessionToken }) {
     this.accessKeyId = accessKeyId
     this.secretAccessKey = secretAccessKey
     /** @type {string} */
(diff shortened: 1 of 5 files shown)


Gozala:
@hugomrdias I don't know why this is still failing, it should be fixed by #67 and it is on my device. I suspect there is some caching happening and new changes in target branch aren't picked up on reruns.

Gozala:
@ice-breaker-tg do you mind doing `git fetch && git merge origin/main && git push` which I believe should fix the failing test. Since pr is from your fork, I'm unable to add any commits here.

ice-breaker-tg:
> @ice-breaker-tg do you mind doing `git fetch && git merge origin/main && git push` which I believe should fix the failing test. Since pr is from your fork, I'm unable to add any commits here.

Did this, ran tests locally and they all passed so hopefully should be good.
Let me know if there is anything else.

w3/sigv4/src/index.js
```diff
@@ -55,6 +56,9 @@ export class SigV4 {
       this.accessKeyId + '/' + this.credentialString
     )
     params.set('X-Amz-SignedHeaders', this.signedHeaders)
+    if (this.securityToken) {
+      params.set('x-amz-security-token', this.securityToken)
```
hugomrdias, line 60:
```suggestion
      params.set('X-Amz-Security-Token', this.securityToken)
```

hugomrdias (changes requested):
Can you please change the param name to `sessionToken` to keep it the same as the s3 sdk ?

hugomrdias (approved):

PR #63, opened 2022-06-26
Fix/cross platform paths

the-simian:
This will break on some operating systems that will not be named so their feelings don't get hurt. the `path` module smooths out inconsistencies like slash direction.


the-simian:
wrong branch.

PR #64, opened 2022-06-26
fix: make paths work on every platform

the-simian:
This will break on some operating systems that will not be named so their feelings don't get hurt. the `path` module smooths out inconsistencies like slash direction.

### DIFF
--- a/w3/cli/src/main.js
+++ b/w3/cli/src/main.js
@@ -16,6 +16,7 @@ import * as Service from 'w3-store'
 import * as CAR from '@ucanto/transport/car'
 import * as FS from 'node:fs/promises'
 import fetch from '@web-std/fetch'
+import path from 'path'
 
 const cli = Soly.createCLI('w3-cli')
 cli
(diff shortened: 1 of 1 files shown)


w3/cli/src/main.js
```diff
@@ -16,6 +16,7 @@ import * as Service from 'w3-store'
 import * as CAR from '@ucanto/transport/car'
 import * as FS from 'node:fs/promises'
 import fetch from '@web-std/fetch'
+import path from 'path'
```
Gozala, line 19:
```suggestion
import path from 'path'
import { pathToFileURL } from 'url'
```

w3/cli/src/main.js
```diff
@@ -361,10 +362,9 @@ const connect = ({
 
 /**
  *
- * @param {string} path
+ * @param {string} relativeFilepath
  */
 
-const resolveURL = (path) =>
-  new URL(path, new URL(`${process.cwd()}`, import.meta.url).href + '/')
+const resolveURL = (relativeFilepath) => path.resolve(`${process.cwd()}`, relativeFilepath)
```
Gozala, line 368:
```suggestion
const resolveURL = (relativeFilepath) =>
  pathToFileURL(path.resolve(process.cwd(), relativeFilepath))
```

Gozala (approved):
Thank you for the fix!

PR #68, opened 2022-06-29
feat: streaming multihasher

Gozala:


w3/streaming-multihash/src/hasher.js
```diff
@@ -0,0 +1,56 @@
...
+  /**
+   * @param {API.AwaitIterable<Uint8Array>} stream
+   */
+  const digestStream = async (stream) => {
+    const hash = hasher.create()
+    for await (const chunk of stream) {
+      hash.update(chunk)
+    }
+    return Digest.create(code, hash.digest())
+  }
```
hugomrdias, line 28: there's a suggestion below to move this to the StreamingHasher class

w3/streaming-multihash/src/hasher.js
```diff
@@ -0,0 +1,56 @@
...
+ * @param {Name} options.name
+ * @param {Code} options.code
+ * @param {{create: () => API.RawHasher}} options.hasher
+ */
+
+export const derive = ({ name, code, hasher }) => {
+  /**
+   * @returns {API.StreamingHasher<Code>}
+   */
+  const create = () => new StreamingHasher(code, hasher.create())
```
hugomrdias, line 17: if you agree with all my suggestions, you dont actually need this you can just return the instance directly

w3/streaming-multihash/src/hasher.js
```diff
@@ -0,0 +1,56 @@
...
+   */
+  const digestStream = async (stream) => {
+    const hash = hasher.create()
+    for await (const chunk of stream) {
+      hash.update(chunk)
+    }
+    return Digest.create(code, hash.digest())
+  }
+
+  return { name, code, create, digestStream }
```
hugomrdias, line 30:
```suggestion
  return { name, code, hasher: new StreamingHasher(code, hasher) }
```

this is all it is, `derives` wraps any hasher in a multihash based hasher with support for `AwaitIterable` directly

w3/streaming-multihash/src/sha256.js
```diff
@@ -0,0 +1,8 @@
+import { sha256 } from '@noble/hashes/sha256'
+import { derive } from './hasher.js'
+
+export const { name, code, create, digestStream } = derive({
+  name: 'sha2-512',
+  code: 0x12,
+  hasher: sha256,
+})
```
hugomrdias, line 8: these wrappers should probably return an fn with the same params as the "create" fn for the hash, call the "create" and send the result (`RawHasher`) to `derive`

https://github.com/paulmillr/noble-hashes#blake2b-blake2s

w3/streaming-multihash/src/hasher.js
hugomrdias: actually if we add the `name` as a property to `StreamingHasher` we can just return `new StreamingHasher`

hugomrdias (changes requested):
sorry theres probably to many suggestions in this review but imo it should simplify and add more flexibility to this

w3/streaming-multihash/src/hasher.js
```diff
@@ -0,0 +1,56 @@
...
+  /**
+   * @param {Uint8Array} bytes
+   */
+  write(bytes) {
+    this.hasher.update(bytes)
+    return this
+  }
+  close() {
+    return Digest.create(this.code, this.hasher.digest())
+  }
```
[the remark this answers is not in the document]
Gozala: This is really confusing name IMO.

w3/streaming-multihash/src/sha256.js
Gozala: Intention here was that `derive` generates `StreamingMultihashHasher` which is roughly an equivalent of [`MultihashHasher`](https://github.com/multiformats/js-multiformats/blob/bb14a29dd823a517ef0c6c741d265e022591d831/src/hashes/interface.ts#L38-L58)

```js
export interface MultihashHasher<Code extends number = number> {
  /**
   * Takes binary `input` and returns it (multi) hash digest. Return value is
   * either promise of a digest or a digest. This way general use can `await`
   * while performance critical code may asses return value to decide whether
   * await is needed.
   *
   * @param {Uint8Array} input
   */
  digest(input: Uint8Array): Promise<MultihashDigest> | MultihashDigest

  /**
   * Name of the multihash
   */
  name: string

  /**
   * Code of the multihash
   */
  code: Code
}
```

w3/streaming-multihash/src/type.ts
```diff
@@ -0,0 +1,21 @@
+import type { MultihashDigest } from 'multiformats/hashes/interface'
+
+export interface StreamingMultihashHasher<Code extends number = number> {
+  readonly code: Code
```
Gozala, line 4:
```suggestion
  /**
   * Name of the multihash
   */
  readonly name: string

  /**
   * Code of the multihash
   */
  readonly code: Code
  
  digestStream(input:AwaitIterable<Uint8Array>): Await<MultihashDigest<Code>>
```

w3/streaming-multihash/src/type.ts
```diff
@@ -0,0 +1,35 @@
...
+   * Name of the multihash
+   */
+  readonly name: string
+
+  /**
+   * Code of the multihash
+   */
+  readonly code: Code
+  
+  **
```
Gozala, line 14:
```suggestion
  /**
```

w3/streaming-multihash/src/type.ts
```diff
@@ -0,0 +1,21 @@
+import type { MultihashDigest } from 'multiformats/hashes/interface'
+
+export interface StreamingMultihashHasher<Code extends number = number> {
+  readonly code: Code
+
+  create(): StreamingHasher<Code>
+}
```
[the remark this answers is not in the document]
Gozala: I have a cli patch that uses this API, I'll submit a separate PR with it.

Gozala:
@hugomrdias It's difficult to convey all the details across multiple comments so I'll try to consolidate all here:

1. I really want `StreamingMultihashHasher` to be pretty much equivalent of `MultihashHasher` from multiformats, with only difference that it operates on an input stream as opposed to in memory buffer.
   - That is also why `StreamingMultihashHasher` has `digestStream` method that is streaming equivalent of `digest` in `MultihashHasher`.
   - It is best to not overload `digest` as that way you could have a modules that implements both interfaces.
2. I have also added `create()` method to `StreamingMultihashHasher` which provides a "writer" interface. Such interfaces are useful when user needs to control stream consumption as opposed to giving that control to the hasher.
    > Good example of this old exporter which takes a stream of files and turns it into UnixFS DAG vs new UnixFS exporter which instead allows you to incrementally build such a DAG. Whenever you need to pause do some things than resume or possibly abort passing streams is not a great API to work with.
    - `StreamingHasher` has methods named `write` and `close` because ideal it would have been Web [WritableStream Writer](https://developer.mozilla.org/en-US/docs/Web/API/WritableStreamDefaultWriter), but given it's browser implementation status it seems better to just have compatible API.
      > It is worth pointing out that same names are used in new UnixFS code as well and for exact same reasons. I hope in the future we would be able to actually replace custom constructs with writable streams.
   - Maybe it should be renamed to `createWriter()` to make this more explicit ?
3. I'm really opposed to idea having `stream(input)` method along the side of `write` / `close` (or suggested `update` / `digest`) because:
   - As elaborated  above you want to either give the control to the resource (your `stream` method or my `digestStream` function) or you want to retain that control. 
   - If you mix both styles together it's unclear who controls the resource. E.g. what happens when you do `hasher.write(bytes).write(bytes).stream(stream)` ? Are first two writes retained or not ? And what happens if you call `.digest()` after `.stream()` was called.
   
   Put it differently `.digestStream()` exposes no state that can be messed with, while `.create()` does the opposite it gives you a state to do whatever you want with it. This makes responsibilities clear.
   > P.S.: This is also how Rust's ownership model guides you to author code


Please let me know if my reasoning resonates and given the context what would you like me to do.

Gozala:
I should extract this into own package

PR #74, opened 2022-07-04
Fix/cross platform urls

the-simian:
Not sure why, there was still a lingering issue with this url resolver for me, but this fixes it.
The main change is removing "pathToFileURL" which produced the error:
```sh
TypeError [ERR_INVALID_URL]: Invalid URL
    at new NodeError (node:internal/errors:371:5)
    at onParseError (node:internal/url:552:9)
    at parse (<anonymous>)
    at new URL (node:internal/url:628:5)
``` 
~~path.resolve by itself seems to work ok~~

It needs to be a URL type for your typing

### DIFF
--- a/w3/cli/src/main.js
+++ b/w3/cli/src/main.js
@@ -17,6 +17,7 @@ import * as CAR from '@ucanto/transport/car'
 import * as FS from 'node:fs/promises'
 import fetch from '@web-std/fetch'
 import path from 'path'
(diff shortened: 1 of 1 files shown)


the-simian:
ok , actually can close this, turns out that what is on npm just doesn't have this fix. I assumed it did so I 'fixed' it and aside from indicating the return type, I've unlocked achievement: fixed the same bug twice in the exact same way.

PR #77, opened 2022-07-11
Pass service DID to accounting

Gozala:
In our last sync call it was decided that we want to track service DID (which is either one for nft.storage or web3.storage) in our accounting. This pull request:

1. Amends `Accounting` to add additional `service` parameter to all methods.
2. Updates `store/*` service implementation so it passes invocation audience DID into `service` parameter.
3. Updates reference implementation and tests accordingly.

@ice-breaker-tg we would need to update actual `Accounting` provider implementation as well, but unlike ref implementation we probably should separate columns for `service` and `group` fields instead of combining combining them together.

@mikeal @hugomrdias tagging you here as well because right now we aren't passing `service` info into `identity/resolve` and we may want such thing if two diverge into separate entities.

### DIFF
--- a/w3/store/src/accounting/lib.js
+++ b/w3/store/src/accounting/lib.js
@@ -1,25 +1,27 @@
 import * as API from '../type.js'
 
 /**
+ * @typdef {[service:DID, group:DID, car:DID, proof:API.Accounting.Link, ]} Table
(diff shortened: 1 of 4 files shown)


Gozala:
@ice-breaker-tg can you please take over this ? I think it's done, but I probably requires some changes on your end so it would be easier for you to just land when it's best.

PR #79, opened 2022-07-27
Pick from access

hugomrdias:


packages/server/src/server.js
```diff
@@ -199,6 +214,32 @@ class HandlerExecutionError extends Error {
...
+    super()
+    /** @type {true} */
+    this.error = true
+    this.caps = caps
+  }
+  get name() {
+    return 'InvocationCapabilityError'
+  }
+  get message() {
+    return `Invocation is required to have one single capability.`
```
Gozala, line 231:
```suggestion
    return `Invocation is required to have a single capability.`
```

packages/server/test/handler.spec.js
```diff
@@ -157,3 +157,53 @@ test('checks service id', async () => {
...
+    can: 'identity/register',
+    with: 'mailto:bob@web.mail',
+  })
+
+  const result = await invocation.execute(client)
+
+  assert.containSubset(result, {
+    error: true,
+    name: 'InvocationCapabilityError',
+    message: 'Invocation is required to have one single capability.',
```
Gozala, line 203:
```suggestion
    message: 'Invocation is required to have a single capability.',
```

packages/core/src/delegation.js
```diff
@@ -121,6 +121,28 @@ export class Delegation {
   get facts() {
     return this.data.facts
   }
+
+  /**
+   * Iterate over the proofs
+   *
+   * @returns {IterableIterator<API.Delegation>}
+   */
+  iterate() {
```
Gozala, line 130: I don't see this been used anywhere but a it's own test, which makes me wonder if we really need it. I'm just hesitant to increasing API surface as it's a lot hard to remove things in the future. Unless we really need it I'd much rather leave this out and maybe export static `iterate` below instead. IMO method only useful if different implementations can have different logic in them, here I can't imagine different implementation of `API.Delegation` that would not work with static `iterate`.

Gozala (approved):
Looks good to me! Suggested few changes but I leave it up to you whether you want to merge as is or address them.

packages/core/src/delegation.js
hugomrdias: i use it in w3access, IMO is nicer to have it in the instance we are working with already than to import another method. the static function is still there for any other use or implementation.

PR #102, opened 2022-09-30
feat: implement rsa signer / verifier

Gozala:
General overview of changes:

1. ucanto now extends `UCAN.Signer` interface with `toArchive(): SignerArchive` method which can be used to save signers across sessions.
2. `SignerArchive` is a type union that represents either binary encoded private key with multiformat code, or `{ did: string, key: CryptoKey }` struct that can be stored in indexdb.
3. ucanto defines new `SignerImporter` interface with `from(archive:SignerArchive): Signer` method, which can be used to import archived signer. In other words `importer.from(signer.toArchive())` roundtrips.
4. similar to `SignerImporter` we have `PrincipalParser` with `parse(did:DID): Verifire` method, which can be used to turn verifiers to DIDs and back `parser.parse(verifier.did())`.
5. Top level `@ucanto/principal` now provides `Signer` export which is composed RSA and Ed `SignerImporter`.
6. `@ucanto/principal` now provides `Verifier` export which is composed RSA and Ed `PrincipalParser`.


packages/principal/src/ed25519/type.ts
```diff
@@ -0,0 +1,26 @@
...
+  encode(): ByteView<EdSigner<M>>
+}
+
+export interface EdVerifier<M extends string = 'key'>
+  extends Verifier<M, CODE> {
+  readonly code: 0xed
+  readonly signatureCode: CODE
+  readonly signatureAlgorithm: ALG
+
+  encode: () => ByteView<EdVerifier<M>>
```
Gozala, line 25:
```suggestion
  encode: () => ByteView<EdVerifier<M>>
  export: () => ByteView<EdVerifier<M>>
```

packages/interface/src/lib.ts
```diff
@@ -388,7 +386,23 @@ export type URI<P extends Protocol = Protocol> = `${P}${string}` &
...
+}
+
+export interface IntoSigner {
+  decode: (bytes: Uint8Array) => Signer
+}
+
+export interface Signer<M extends string = string, A extends number = number>
+  extends UCANSigner<M, A> {
+  export?: () => Await<ByteView<Signer<M, A>>>
+  toCryptoKey?: () => Await<CryptoKey>
```
hugomrdias, line 399:
```suggestion
  exportBytes?: () => Await<ByteView<Signer<M, A>>>
  exportKey?: () => Await<CryptoKey>
```

packages/interface/src/lib.ts
```diff
@@ -388,7 +386,23 @@ export type URI<P extends Protocol = Protocol> = `${P}${string}` &
...
+}
+
+export interface Verifier<M extends string = string, A extends number = number>
+  extends UCANVerifier<M, A> {
+  export?: () => Await<ByteView<Verifier<M, A>>>
+  toCryptoKey?: () => Await<CryptoKey>
```
hugomrdias, line 405:
```suggestion
  exportBytes?: () => Await<ByteView<Verifier<M, A>>>
  exportKey?: () => Await<CryptoKey>
```

packages/principal/test/ed25519.spec.js
```diff
@@ -0,0 +1,181 @@
...
+    assert.equal(await signer.verify(payload, signature), true)
+
+    assert.equal(signer.signatureAlgorithm, 'EdDSA')
+    assert.equal(signer.signatureCode, 0xd0ed)
+  })
+
+  it('derive', async () => {
+    const original = await Lib.generate()
+    // @ts-expect-error - secret is not defined by interface
+    const derived = await Lib.derive(original.secret)
```
hugomrdias, line 50: ed has a derive method ? we need to have a smaller api surface please and consistency in the names what ever they are

hugomrdias (changes requested):

Gozala:
@hugomrdias I have trimmed interface per your feedback. Now `Verifier` is simply a `UCAN.Verifier` and `Signer` only extends `UCAN.Signer` with a single `toArchive` method which returns either bytes (if underlying key is extractable) or `{did, key}` pair which can be saved into IndexedDB and later restored via `Signer.from(archive)`. This should remove any need for try catch / or if guarding.

I hope these changes address your concerns, please have another pass and let me know whether to proceed or if there are some other concerns to be addressed.

packages/interface/src/lib.ts
```diff
@@ -388,9 +386,61 @@ export type URI<P extends Protocol = Protocol> = `${P}${string}` &
...
+   * Returns archive of this signer which is byte encoded form when signer key
+   * is extractable and is {@link SignerInfo} form otherwise. This allows a user
+   * to store non extractable archives in indexedDB and store extractable
+   * archives on disk, which matches general expectation that in browsers
+   * unextratable keys should be used and extractable keys in node.
+   *
+   * @example
+   * ```ts
+   * const save = async (signer: Signer) => {
+   *   const archive = signer.toAchive()
```
hugomrdias, line 415:
```suggestion
   *   const archive = signer.toArchive()
```

packages/interface/src/lib.ts
```diff
@@ -388,9 +386,61 @@ export type URI<P extends Protocol = Protocol> = `${P}${string}` &
...
+   *   if (archive instanceof Uint8Array) {
+   *     await fs.writeFile(KEY_PATH, archive)
+   *   } else (signer.exportKey) {
```
hugomrdias, line 418:
```suggestion
   *   } else {
```

packages/principal/src/rsa.js
```diff
@@ -0,0 +1,322 @@
...
+  verify(payload, signature) {
+    return this.verifier.verify(payload, signature)
+  }
+  /**
+   * @template T
+   * @param {API.ByteView<T>} payload
+   * @returns {Promise<API.Signature<T, typeof signatureCode>>}
+   */
+  async sign(payload) {
+    this.privateKey
```
hugomrdias, line 273:
```suggestion
```

packages/principal/src/rsa/private-key.js
```diff
@@ -0,0 +1,127 @@
...
+ * @param {RSAPrivateKey} key
+ */
+export const toSPKI = key => SPKI.encode(PublicKey.encode(key))
+
+/**
+ *
+ * @param {string|undefined} input
+ * @returns
+ */
+const base6urlDecode = (input = '') => base64url.baseDecode(input)
```
hugomrdias, line 127:
```suggestion
const base64urlDecode = (input = '') => base64url.baseDecode(input)
```

hugomrdias (approved):
LGTM

PR #104, opened 2022-09-30
feat: only use native fetch type and Headers object

hugomrdias:
avoid ts errors when passing fetch implementation and normally we only have Headers object in all the env so no need to deal we other types


Gozala (changes requested):
Hey @hugomrdias I'm afraid `Headers` aren't a thing even in node 16 which is current TLS version, which is also why I originally not used them.

Can we address original issue without breaking node 16 support or introducing dependence on an external library ?

Gozala:
Problem this was addressing is fixed by #113, we can revisit this after node is 18 TLS and available on AWS

PR #109, opened 2022-10-11
feat: rip out special handling of my: and as: capabilities

Gozala:
This remove built-in support for my: and and as: capability schemas that were removed in 0.9 UCAN spec.

PR targets  [feat/decoders-dsl](https://github.com/web3-storage/ucanto/tree/feat/decoders-dsl) branch because it has better test coverage and targeting main would introduce more conflicts.


hugomrdias (approved):

packages/validator/test/extra-schema.spec.js
```diff
@@ -169,7 +169,7 @@ test('URI.from', () => {
     })
 
     test(`Link.optional().read(${input})`, () => {
-      const link = Link.optional()
+      const link = Link.link().optional()
```
Gozala, line 172:
```suggestion
      const link = Link.optional()
```

PR #122, opened 2022-11-09
chore: add doc comments to interface/lib.ts

yusefnapora:
I started reading a bit deeper into ucanto as I'm thinking about ways to document the capability and service definitions.

I added some doc comments to some of the types in the main `interface` package, mostly to try to clarify my understanding. I haven't got to the more gnarly inference / capability parsing stuff yet and may leave that stuff for later. Mostly just want to cover types that you might encounter if you're defining a ucanto service implementation.

@Gozala could you look through and see if I'm describing things correctly?

### DIFF
--- a/packages/interface/src/lib.ts
+++ b/packages/interface/src/lib.ts
@@ -170,13 +170,7 @@ export interface Delegation<C extends Capabilities = Capabilities> {
  * sufficient privileges to execute.
  */
 export interface Invocation<C extends Capability = Capability>
-  extends Delegation<[C]> {
-  /** The `issuer` for an {@link Invocation} is the party requesting the invocation from the service provider. */
-  issuer: Signer
-
-  /** The `audience` for an {@link Invocation} is the service provider who will execute the invocation. */
-  audience: Principal
-}
+  extends Delegation<[C]> {}
 
 /**
  * A {@link UCANOptions} instance that includes options specific to {@link Invocation}s.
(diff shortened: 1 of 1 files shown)


packages/interface/src/lib.ts
```diff
@@ -369,29 +448,57 @@ export interface Server<T> extends ServerOptions {
...
 
+/**
+ * A string literal type that matches the "scheme" portion of a URI.
+ */
 export type Protocol<Scheme extends string = string> = `${Scheme}:`
 
+/**
+ * A typed string representing a URI of a given protocol.
+ *
+ * @typeParam P - The protocol (scheme) of the given uri. For example, `did:key:foo` has the protocol of `did`.
```
Gozala, line 489:
```suggestion
 * @template P - The protocol (scheme) of the given uri. For example, `did:key:foo` has the protocol of `did`.
```

Gozala:
Thank you and sorry for not getting to this sooner.

packages/interface/src/lib.ts
```diff
@@ -123,12 +158,35 @@ export interface Delegation<C extends Capabilities = Capabilities> {
...
+ */
 export interface Invocation<C extends Capability = Capability>
-  extends Delegation<[C]> {}
+  extends Delegation<[C]> {
+  /** The `issuer` for an {@link Invocation} is the party requesting the invocation from the service provider. */
+  issuer: Signer
+
+  /** The `audience` for an {@link Invocation} is the service provider who will execute the invocation. */
+  audience: Principal
+}
```
Gozala, line 179:
```suggestion
  extends Delegation<[C]> {}
```

PR #124, opened 2022-11-10
fix: optional field validation

### LINKED ISSUE #123: Validator fails if some `nb` fields are optional & omitted in delegations
Here is an example of the issue I've encountered in the `w3-protocol` repo:

```js
const any = async () =>
  Capability.any.delegate({
    issuer: account,
    audience: alice,
    with: account.did(),
  })


const delegation = await Store.add.delegate({
  issuer: alice,
  audience: bob,
  with: account.did(),
  nb: {
    size: 1024,
  },
  proofs: [await any()],
})

const add = Store.add.invoke({
  issuer: bob,
  audience: w3,
  with: account.did(),
  nb: {
    size: 1000,
    link: parseLink('bafkqaaa'),
  },
  proofs: [delegation],
})

const result = await access(await add.delegate(), {
  capability: Store.add,
  principal: Verifier,
})
```

It fails with following error

```
AssertionError [ERR_ASSERTION]: Claimed capability {"can":"store/add","with":"did:key:z6MktafZTREjJkvV5mfJxcLpNBoVPwDLhTuMg9ng7dY4zMAL","nb":{"link":{"code":85,"version":1,"hash":{"0":0,"1":0}},"size":1000}} is invalid
  - Capability can not be (self) issued by 'did:key:z6MkffDZCkCTWreg8868fG1FGFogcJj5X6PY93pPcWDn9bob'
  - Can not derive {"can":"store/add","with":"did:key:z6MktafZTREjJkvV5mfJxcLpNBoVPwDLhTuMg9ng7dY4zMAL","nb":{"link":{"code":85,"version":1,"hash":{"0":0,"1":0}},"size":1000}} from delegated capabilities:
    - Encountered malformed 'store/add' capability: {"nb":{"size":1024},"can":"store/add","with":"did:key:z6MktafZTREjJkvV5mfJxcLpNBoVPwDLhTuMg9ng7dY4zMAL"}
      - Expected link but got undefined instead
```

Which happens because `delegation` contains no `link` & I'm guessing validator parses capability not as delegation but as an invocation.

Gozala:
Fixes #123 

### Overview

- [x] Adds `optional` parameter to the `parse` function used by validator to parse capabilities.
- [x] `.select` method which selects matching proofs from delegation chain passes `option:true` to parser so `nb` fields are never required in delegations.
- [x] `derives` of the capability infers type `nb` fields as optionals.
- [x] `derives` in `cap.derive({ to, derives })` infers `nb` fields as optionals.
- [x] `.delegate(opts).capabilities[0].nb` infers fields as optionals.
- [x] `.invoke(opts).capabilities[0].nb` infers fields as defined in schema.

---

I hate introducing more utility types to make this work, but unfortunately there was no simple way to fix all the problems. It is possible to rework validator such that we would be able to remove bunch of these utility types, however it is non trivial change and I'd rather do it later on when there's less urgency.

### DIFF
--- a/packages/validator/src/capability.js
+++ b/packages/validator/src/capability.js
@@ -432,7 +432,7 @@ class Match {
     const errors = []
     const matches = []
     for (const capability of capabilities) {
-      const result = parse(this, capability)
+      const result = parse(this, capability, true)
       if (!result.error) {
         const claim = this.descriptor.derives(this.value, result)
         if (claim.error) {
@@ -640,10 +640,11 @@ class AndMatch {
  * @template {API.Caveats} C
  * @param {{descriptor: API.Descriptor<A, R, C>}} self
  * @param {API.Source} source
+ * @param {boolean} optional
  * @returns {API.Result<API.ParsedCapability<A, R, API.InferCaveats<C>>, API.InvalidCapability>}
  */
 
-const parse = (self, source) => {
+const parse = (self, source, optional = false) => {
   const { can, with: withReader, nb: readers } = self.descriptor
   const { delegation } = source
   const capability = /** @type {API.Capability<A, R, API.InferCaveats<C>>} */ (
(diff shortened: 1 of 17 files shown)


packages/validator/test/inference.spec.js
```diff
@@ -93,3 +93,152 @@ test('use InferInvokedCapability', () =>
...
+test('infers nb fields optional', () => {
+  capability({
+    can: 'test/nb',
+    with: DID.match({ method: 'key' }),
+    nb: {
+      msg: URI.match({ protocol: 'data:' }),
+    },
+    derives: (claim, proof) => {
+      /** @type {string} */
+      // @ts-expect-error - may be undenfined
```
alanshaw, line 106: This typo occurs many times...

```suggestion
      // @ts-expect-error - may be undefined
```

packages/validator/src/capability.js
```diff
@@ -640,10 +640,11 @@ class AndMatch {
  * @template {API.Caveats} C
  * @param {{descriptor: API.Descriptor<A, R, C>}} self
  * @param {API.Source} source
+ * @param {boolean} optional
```
alanshaw, line 643: Please can we get a comment to specify what `optional` means while parsing?

alanshaw (approved):
This looks ok to me, although with the caveat that I don't have a a very deep level of understanding of ucanto yet...

PR #139, opened 2022-11-18
feat: did:dns support

### LINKED ISSUE #137: Add `did:dns` support
This is intended to support https://github.com/web3-storage/w3protocol/issues/182 use case

Gozala:
Fixes #137

Adds a new DNS principal which:

1. Can be a simple wrapper around other principals
2. Who's `Validator` can  be passed custom `resolve` handler to allow key resolution via DNSLookup or other means.

### DIFF
--- a/packages/interface/src/lib.ts
+++ b/packages/interface/src/lib.ts
@@ -422,11 +422,19 @@ export interface Signer<M extends string = string, A extends number = number>
    * }
    * ```
    */
-  toArchive(): SignerArchive<Signer<M, A>>
+  toArchive(): SignerArchive<this>
 }
 
 export interface SignerInfo<Self extends Signer = Signer> {
+  /**
+   * DID that is returned by .did() method of the signer.
+   */
   readonly did: ReturnType<Self['did']>
+  /**
+   * If .did() returns non did:key (e.g did:dns) this field will
+   * contain did:key it resolves to, otherwise it is omitted.
+   */
+  readonly resolvedDID?: DID<'key'>
   readonly key: CryptoKey
 }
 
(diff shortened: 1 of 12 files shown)


alanshaw (approved):

packages/interface/src/lib.ts
```diff
@@ -422,11 +422,19 @@ export interface Signer<M extends string = string, A extends number = number>
...
 export interface SignerInfo<Self extends Signer = Signer> {
+  /**
+   * DID that is returned by .did() method of the signer.
+   */
   readonly did: ReturnType<Self['did']>
+  /**
+   * If .did() returns non did:key (e.g did:dns) this field will
+   * contain did:key it resolves to, otherwise it is omitted.
+   */
+  readonly resolvedDID?: DID<'key'>
```
gobengo, line 437: This is probably fine for now, but in general I would expect a `resolveDID` operation to return a [DID Document]() type and not a `did:key`. A did document can have many [verificationMethods](https://www.w3.org/TR/did-core/#verification-methods) (e.g. key) and each vm can be authorized to be used only for certain [verification relationships](https://www.w3.org/TR/did-core/#verification-relationships), e.g. authentication vs capabilityInvocation. 

It's only `did:key` and one or two others that have a 1:1 correspondence between did and key, and where the key is automatically usable for all the verificaiton relationships.

However, I don't see any normative text in did-core that really requires anyone to be conscienscious of this... and did:dns and did:web dont either afaict (see this did:dns issue https://github.com/danubetech/did-method-dns/issues/7 )

gobengo: A more pressing issue is that the way I read `did:dns`, a `did:dns` could correspond to more than 1 `did:key` (regardless of verificationRelationships). https://danubetech.github.io/did-method-dns/#resolve

Probably should consider what will happen if there is more than one key. IMO we shouldn't just 'pick one' but should have this `resolveDID` support returning a `Set<Did<'key'>>`

Gozala:
Closing in favor of https://github.com/web3-storage/ucanto/pull/149

PR #168, opened 2022-12-08
feat: embedded key resolution

Gozala:
Implements embedded key resolution non did:key principals as per https://github.com/web3-storage/specs/blob/main/w3-account.md#update


packages/interface/src/capability.ts
```diff
@@ -358,18 +362,36 @@ export interface ProofResolver extends PrincipalOptions {
...
 }
 
-export interface ValidationOptions<C extends ParsedCapability>
-  extends Partial<CanIssue>,
+export interface Docs {
+  [key: DID]: { key: DIDKey }
+}
+
+export interface Validator {
+  authority: Verifier
```
gobengo, line 375: Could be helpful to jsdoc what `authority` means in this context

packages/interface/src/capability.ts
```diff
@@ -405,6 +427,13 @@ export interface UnavailableProof extends Failure {
   readonly link: UCANLink
 }
 
+export interface DIDResolutionError extends Failure {
+  readonly name: 'DIDResolutionError'
+  readonly did: UCAN.DID
+
+  readonly cause?: Unauthorized
```
gobengo, line 434: Is it really important that the `cause` always be `Unauthorized` and not another Error?

packages/interface/src/lib.ts
```diff
@@ -499,6 +504,10 @@ export interface PrincipalParser {
   parse(did: UCAN.DID): Verifier
 }
 
+export interface DIDResolver {
+  resolveDID?: (did: UCAN.DID) => Await<Result<DIDKey, DIDResolutionError>>
```
gobengo, line 508: For the long term, I bristle a bit seeing the only thing that a `UCAN.DID` can resolve to is a single `DIDKey`.

What if we had 
```
resolveDID<DidMethod>?: (did: UCAN.DID<DidMethod>) => Await<Result<DIDResolution<DidMethod>, DIDResolutionError>>
```

and

```
type SingleKeypairDidDocument<ID extends UCAN.DID> = {
  id: ID,
  /** https://www.w3.org/TR/did-core/#also-known-as */
  alsoKnownAs: DIDKey
}
type DidResolution<ID> =
| KeyAliasDidDocument<ID>
```

gobengo: alternatively could have `resolveDID` -> `resolveDIDKey`.

I like that it gets nowhere near 'did resolution' nor reinforces the (IMO footgun) assumption that a DID will always cleanly resolve to a single default keypair.

gobengo: some more on did resolution result https://w3c-ccg.github.io/did-resolution/#did-resolution-result

gobengo: and impl https://github.com/decentralized-identity/did-resolver/blob/master/src/resolver.ts#L27

packages/principal/test/ed25519.spec.js
```diff
@@ -6,6 +6,8 @@ import { varint } from 'multiformats'
 describe('signing principal', () => {
   const { Signer } = Lib
 
+  /** @type {Lib.Signer.EdSigner} */
```
gobengo, line 9: Is this type assertion doin anything? I'm not sure what line it asserts about given the newline before and after

packages/validator/src/lib.js
```diff
@@ -143,37 +154,83 @@ const isSelfIssued = (capability, issuer) => capability.with === issuer
  * @param {API.ValidationOptions<API.ParsedCapability<A, R, API.InferCaveats<C>>>} config
  * @returns {Promise<API.Result<Authorization<API.ParsedCapability<A, R, API.InferCaveats<C>>>, API.Unauthorized>>}
  */
-export const access = async (
-  invocation,
-  { canIssue = isSelfIssued, principal, resolve = unavailable, capability }
+export const access = async (invocation, { capability, ...config }) =>
```
gobengo, line 157: I think a human readable js doc explaining when/why to use this could be helpful. In general, I like short names like these packages use, but we also frequently re-use those names across modules/files, and there are a lot of objects whose property names come from the intersection of several types, and those properties sometimes come from exports like this via a glob import.

wdyt of jsdoc human readable explanation here?
wdyt of something like an eslint rule that requires human readable explainers for exported functions?

packages/validator/src/lib.js
```diff
@@ -320,30 +373,64 @@ class InvalidClaim extends Failure {
 /**
  * @implements {API.Unauthorized}
```
gobengo, line 374: I think you can have tsc enforce this by adding to the constructor last line:
```
/** @type {API.Unauthorized} */ (this);
```

packages/validator/test/mailto.spec.js
```diff
@@ -0,0 +1,195 @@
...
+import { ed25519, Verifier } from '@ucanto/principal'
+import * as Client from '@ucanto/client'
+import * as Core from '@ucanto/core'
+
+import { alice, bob, mallory, service } from './fixtures.js'
+const w3 = service.withDID('did:web:web3.storage')
+
+const claim = capability({
+  can: 'access/claim',
+  with: DID.match({ method: 'mailto' }),
```
gobengo, line 14: I really don't think we should use `did:mailto` if we don't have to. And I think it's a good sign we don't have to if no one wants to know how to go from `did:mailto` to a full did document. iirc in another thread, I pointed this out, and @Gozala seemed open to just using `mailto:` URIs.

Can we just use mailto uris?

If we really need a did, can we prototype using:
`did:web:web3.storage:account:${encodeURIComponent(`mailto:ben@dag.house`)}}` 

(but IMO we don't need a DID at all here)

packages/interface/src/lib.ts
Gozala: You are right `resolveDID` is misleading name, this is not a function to resolve a DID document but rather to resolve a key. I'll rename both interface and a function.

packages/validator/src/lib.js
Gozala: @gobengo this has came up on several of your reviews so lets make sure we're on the same page on this:

If you are under wrong impression that TS does not [enforces `@implements` directive](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#implements), your information may be out of date.

If there is another reason that I fail to recognize for doing type cast please elaborate

packages/validator/test/mailto.spec.js
Gozala: > I think it's a good sign we don't have to if no one wants to know how to go from did:mailto to a full did document.

We do not want that however, what we want to do is find a proof that we (trusted authority) have authorized user agent identified by some did:key to act on behalf of `did:mailto`'s authority.

It does not mean we resolved DID key or DID document (perhaps we did during authentication), we simply set a cookie in the authenticated user agent so we don't have to reauth as long as our authorization is valid.

Gozala: Created an issue to discuss this further https://github.com/web3-storage/specs/issues/25

packages/validator/src/lib.js
gobengo: You're right! Sorry for my confusion. I verified that `@implements` does seem to work with latest. I'm confused by https://github.com/microsoft/TypeScript/issues/30156#issuecomment-952667829

gobengo (approved):

PR #173, opened 2022-12-12
feat: implement toJSON on delegations

### LINKED ISSUE #172: Add toJSON method to delegations

Gozala:
Fixes #172

### DIFF
--- a/packages/core/src/delegation.js
+++ b/packages/core/src/delegation.js
@@ -1,6 +1,7 @@
 import * as UCAN from '@ipld/dag-ucan'
 import * as API from '@ucanto/interface'
 import * as Link from './link.js'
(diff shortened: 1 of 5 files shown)


alanshaw:
If we get a `fromJSON` we can remove some code for storing delegations in the access Conf store.

alanshaw (changes requested):
It's _very_ unexpected for `toJSON` to be lossy. I wouldn't have suggested `fromJSON` if I'd known that was the case.

It either needs to be commented heavily or renamed. I'm not even sure if commenting is enough since `toJSON` is called transparently by `JSON.stringify`.

vasco-santos:
Let's discuss this further once we get first prod release in place

PR #185, opened 2022-12-14
fix: toJSON behavior on the ucan.data

Gozala:
Propagates changes form https://github.com/ipld/js-dag-ucan/pull/72 into ucanto and adds some tests to ensure no regressions

Supersedes #173


vasco-santos (approved):
LGTM

PR #186, opened 2022-12-14
feat: delegation.toJSON

### LINKED ISSUE #172: Add toJSON method to delegations

Gozala:
This supersedes #173 and Fixes #172. It targets #185 as it relies on changes there.

In the other PR @alanshaw had this suggestion

> If we get a fromJSON we can remove some code for storing delegations in the access Conf store.

And this PR still does not add `fromJSON` because, even though it will serialize the whole chain into a JSON which could be then de-serialized into compatible chain there is still a problem with the fact that `fromJSON` would not know if inline proofs do match claimed CIDs. This means that motivated attacker could use this an a attack vector.

In other words code that imports delegation should also validate hashes which is async operation and there for does not fit into a `fromJSON` signature.

That said I think it still useful to have `toJSON` method for debugging purposes, which is what this PR adds.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   "devDependencies": {
     "mocha": "^10.1.0",
     "prettier": "2.7.1",
-    "typescript": "4.8.3"
+    "typescript": "^4.8.4"
   },
   "prettier": {
     "trailingComma": "es5",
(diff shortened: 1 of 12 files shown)


gobengo (approved):
awesome

alanshaw (approved):

vasco-santos (approved):
lgtm thanks @Gozala

Issue #175, opened 2022-12-13
Expose DID.from function


Gozala:
There has been multiple instances of `DID.match({}).from(did)` where `.match({})` seems redundant. We probably should export `from` function to remove a need for that redundant `match`

gobengo:
not urgent

PR #191, opened 2022-12-16
Add @ucanto/validator DID.from

gobengo:
Motivation:
* https://github.com/web3-storage/ucanto/issues/175

### DIFF
--- a/packages/validator/test/extra-schema.spec.js
+++ b/packages/validator/test/extra-schema.spec.js
@@ -381,3 +381,41 @@ test('URI.from', () => {
     })
   }
 }
+
+{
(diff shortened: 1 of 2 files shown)


packages/validator/test/extra-schema.spec.js
```diff
@@ -381,3 +381,41 @@ test('URI.from', () => {
...
+        name: 'TypeError',
+        message: 'Expected value of type string instead got undefined',
+      },
+    ],
+  ]
+  for (const [did, errorExpectation] of dataset) {
+    test(`DID.from("${did}")`, () => {
+      let error
+      try {
+        DID.from(did)
```
alanshaw, line 410: Assert that it returns an expected value?

PR #199, opened 2023-01-20
feat: support execution of delegations

### LINKED ISSUE #195: Connection should be able to take materialized invocations
I was telling @gobengo that we could route request from one server to the other simply by passing them to `connection.execute`, but turns out this was incorrect because:

1. Connection's execute takes `IssuedInvocation` (Well tuple of `ServiceInvocation`s but those are basically `IssuedInvocation`s) which is not an `Delegation` it's like a lazy version that can be materialized as needed

   https://github.com/web3-storage/ucanto/blob/582c4c504d2e75feee2c5d298ea08b2f01ff1c5e/packages/interface/src/lib.ts#L232-L239

2. Connection passes these lazy invocations into encoder to encode

   https://github.com/web3-storage/ucanto/blob/582c4c504d2e75feee2c5d298ea08b2f01ff1c5e/packages/client/src/connection.js#L47-L49

3. Which in turn create actual delegations by materializing them

   https://github.com/web3-storage/ucanto/blob/582c4c504d2e75feee2c5d298ea08b2f01ff1c5e/packages/transport/src/car.js#L19-L23 

Few things to note here:

1. For a while I have been considering to get rid off these lazy invocation things which would make .invoke and .delegate functions more similar. I'm still hesitant however as they introduce additional awaits, first you await to create invocation then you await to execute the invocation.
2. Even with lazy invocations I think it is a mistake to make materializing them transport's problem. It would make a lot more sense to make `execute` materialize all of the invocations first and then pass those onto transport to just do encoding.
   - It probably ended up in transport because there was a loop iterating invocations and I just avoided second one

### Option 1

- I think we should change transport API so it expects materialized invocations and make materializing them connection's problem
- We could also loosen up execute function to allow passing both materalized and non-materialized invocations so it could materialize ones that aren't yet and forward them onto transport

### Option 2

Perhaps whole materialized / not-yet-materalized invocations could be abstracted away with some `EncodableInvocation` interface of sorts that provides some method e.g. `.delegate()` which returns a materialized invocation. That way connection, transport and all those layers could just expect `EncodableInvocation` and not care which one they're actually getting.

Other than type changes it would basically mean adding `delegate() { return this }` to the `Delegation` class because `Invocation` already has `.delegate()`

Gozala:
Fixes #195

### DIFF
--- a/packages/transport/test/car.spec.js
+++ b/packages/transport/test/car.spec.js
@@ -1,7 +1,14 @@
 import { test, assert } from './test.js'
 import * as CAR from '../src/car.js'
 import * as CBOR from '../src/cbor.js'
-import { delegate, Delegation, UCAN, parseLink, isLink } from '@ucanto/core'
+import {
+  delegate,
+  invoke,
+  Delegation,
+  UCAN,
+  parseLink,
+  isLink,
+} from '@ucanto/core'
 import * as UTF8 from '../src/utf8.js'
 import { alice, bob, mallory, service } from './fixtures.js'
 import { CarReader } from '@ipld/car/reader'
@@ -15,18 +22,16 @@ test('encode / decode', async () => {
   const expiration = 1654298135
 
   const request = await CAR.encode([
-    {
+    invoke({
       issuer: alice,
       audience: bob,
-      capabilities: [
-        {
-          can: 'store/add',
-          with: alice.did(),
-        },
-      ],
+      capability: {
+        can: 'store/add',
+        with: alice.did(),
+      },
       expiration,
       proofs: [],
-    },
+    }),
   ])
 
   assert.deepEqual(request.headers, {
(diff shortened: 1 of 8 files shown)


gobengo (approved):

gobengo (approved):

packages/transport/test/car.spec.js
```diff
@@ -121,6 +122,8 @@ test('accepts Content-Type as well', async () => {
     expiration,
   })
 
+  CAR.encode([delegation])
```
gobengo, line 125: is this just asserting that `encode` doesn't throw? or does the call to `encode` have an important side effect?

Gozala: No just a mistake, I'll remove it

hugomrdias:
@Gozala where is this https://github.com/web3-storage/ucanto/blob/e84c659268d104e24bfc0d0354ef436b865b12ca/packages/core/src/delegation.js#L192 function called now ? This is where the ucan is signed and it does not seems to be called.

gobengo:
@hugomrdias that `delegate` fn is called by `IssuedInvocation#delegate` https://github.com/web3-storage/ucanto/blob/e84c659268d104e24bfc0d0354ef436b865b12ca/packages/core/src/invocation.js#L52 which, after this PR, is called by the transport encoders [eg here](https://github.com/web3-storage/ucanto/pull/199/files#diff-520f4a0baea89881cb7747d65a4ad3af8144dab526c7edd4d9597995c73eb5e2R23)

PR #212, opened 2023-01-31
readme updates

jchris:
Let's come back to this after w3up release, but these are some places the README could use cleanup.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -91,13 +91,13 @@ The library comes with several transport layer codecs you can pick from, but you
 import * as Server from "@ucanto/server"
 import * as CAR from "@ucanto/transport/car"
 import * as CBOR from "@ucanto/transport/cbor"
-import { SigningPrincipal } from "@ucanto/principal"
+import { ed25519 } from "@ucanto/principal"
 import * as HTTP from "node:http"
 import * as Buffer from "node:buffer"
 
 export const server = (context { store = new Map() } : { store: Map<string, string> }) =>
   Server.create({
-    id: await SigningPrincipal.derive(process.env.SERVICE_SECRET),
+    id: ed25519.Signer.parse(process.env.SERVICE_SECRET),
     service: service(context),
     decoder: CAR,
     encoder: CBOR,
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -191,9 +190,7 @@ const connection = Client.connect({
 In practice you probably would want client/server communication to happen across the wire, or at least across processes. You can bring your own transport channel, or choose an existing one. For example:
 
 ```ts
-import * as HTTP from '@ucanto/transport/http'
-import * as CAR from '@ucanto/transport/car'
-import * as CBOR from '@ucanto/transport/cbor'
+import Transport from '@ucanto/transport'
```
jchris, line 193: This would require making the Transport directory have an index.js

Readme.md
```diff
@@ -165,7 +164,7 @@ const demo1 = async connection => {
     },
   })
 
-  const result = await me.execute(connection)
+  const result = await connection.execute([me])
```
gobengo, line 167: I'm a little surprised that this would be necessary. I'd expect the former to work?

and it's a little bit more sugary. `connection.execute` handle an array of invocations, but the `me.execute` only executes the `me` invocation (api doesn't involve any array)

gobengo (approved):

Readme.md
gobengo: it would also require it to have a `default` export

Readme.md
jchris: it was crashing the server with a car parse error

gobengo (changes requested):
IMO we could merge this if it had everything except the proposed change to how `@ucanto/transport` does exports (which we can consider separately from fixing [this](https://github.com/web3-storage/ucanto/pull/212/files#diff-1550ec65ac92f65817fc28928dfef526912b5f52356ff43651369bae92f56031L213))

jchris:
ok I that does it -- look ready to merge?

Readme.md
gobengo: @jchris I'll try to watch for that, but if you repro it again, send me the error stack trace

gobengo: example of how we use `IssuedInvocation#execute` in the wild https://github.com/web3-storage/w3protocol/blob/fc53a16c40f1558b2b17dd4ff9e301a1ef54950f/packages/access-api/test/space-info.test.js#L27

gobengo (approved):

gobengo:
Thank you for reporting and fixing this :)

PR #214, opened 2023-02-02
fix: expose url in channel interface

hugomrdias:
Its useful to be able to get the url from the channel directly, in situations where the connection is passed into a function or class and we want to use the connection url has the default for any other communications like WS.

Instead of input url + connection we can just ask the user for connection.

example :  https://github.com/web3-storage/w3protocol/pull/404

### DIFF
--- a/packages/interface/src/transport.ts
+++ b/packages/interface/src/transport.ts
@@ -26,6 +26,7 @@ export interface Channel<T extends Record<string, any>> extends Phantom<T> {
   request<I extends Tuple<ServiceInvocation<UCAN.Capability, T>>>(
     request: HTTPRequest<I>
   ): Await<HTTPResponse<InferServiceInvocations<I, T>>>
+  url: URL
 }
 
 export interface RequestEncoder {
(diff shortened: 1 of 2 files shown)


travis:
reopening, I'm going to take over getting https://github.com/web3-storage/w3protocol/pull/404/ merged - @Gozala would love a review from you when you get a chance!

alanshaw:
The HTTP channel already has a `url` property: https://github.com/web3-storage/ucanto/blob/main/packages/transport/src/http.js#L47

I think you just need an `instanceof` check:

```js
import { Channel as HttpChannel } from '@ucanto/transport/http'

const { channel } = connection
if (!(channel instanceof HttpChannel)) throw new Error('HTTP channel is required')
const { url } = channel
// do something with URL
```

I assume the idea is that not all channels will be HTTP in the future...

If you add the type (this PR) then you're still going to need the conditional to check for existence of the `url` property so you may as well just check with `instanceof` or duck type it.

travis:
closing in favor of a solution to the underlying issue proposed in https://github.com/web3-storage/w3protocol/pull/440

PR #215, opened 2023-02-02
fix: add @returns to @ucanto/validator Schema.dictionary() to avoid problematic inference when using derives

gobengo:
Motivation:
* obviate https://github.com/web3-storage/w3protocol/pull/420#discussion_r1095096263

### DIFF
--- a/packages/validator/test/capability.spec.js
+++ b/packages/validator/test/capability.spec.js
@@ -741,40 +741,6 @@ test('capability amplification', () => {
   )
 })
 
-test('can create derived capability with dict schema in nb', () => {
-  /**
-   * @param {{ with: string }} claim
-   * @param {{ with: string }} proof
-   */
-  const equalWith = (claim, proof) => claim.with === proof.with || new Failure(`claim.with is not proven`);
-  const top = capability({
-    can: '*',
-    with: URI.match({ protocol: 'did:' }),
-    derives: equalWith,
-  });
-  const delegate = top.derive({
-    to: capability({
-      can: 'access/delegate',
-      with: URI,
-      nb: {
-        delegations: Schema.dictionary({
-          value: Schema.Link.match(),
-        }),
-      },
-      derives: (claim, proof) => {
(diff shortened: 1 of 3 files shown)


packages/validator/test/capability.spec.js
```diff
@@ -741,6 +741,40 @@ test('capability amplification', () => {
...
+      with: URI,
+      nb: {
+        delegations: Schema.dictionary({
+          value: Schema.Link.match(),
+        }),
+      },
+      derives: (claim, proof) => {
+        // the motivation for this test was that tsc would previously complain at these assignments
+        // and the only workaround was a type assertion https://github.com/web3-storage/w3protocol/pull/420/commits/4f1f2931cecff1d1d1d29e889c4fdfb63ff3b327#diff-e434cc6c1a699df311a0b2faed199a2ff6b6b291d30f95e20b2ea5abfa7da3d9R125
+        /** @type {Schema.Dictionary|undefined} */
```
gobengo, line 767: `tsc` will complain here without the change adding `@returns` to `Schema.dictionary()`.

Gozala: This is cool, but since we're testing inference as opposed to something at runtime I would suggest putting it in this file instead https://github.com/web3-storage/ucanto/blob/main/packages/validator/test/inference.spec.js

Gozala (approved):
Thanks, minor suggestion but I'm ok with this being as is.

packages/validator/test/capability.spec.js
gobengo: thanks, I did feel weird that it was a tsc-only test alongside a bunch of runtime ones, but figured you'd either tell me to just omit the test or lmk a good place to put it :) 🙏

gobengo: https://github.com/web3-storage/ucanto/pull/215/commits/aef17d94d3b5e7775622b13148295effd23cfa57

PR #218, opened 2023-02-08
feat!: wildcard support in capabilies

### LINKED ISSUE #87: Support `can: store/*` like patterns in validators
Right now validator would not allow you to derive `store/add` from `store/*` because it does not know `store/*` capability. We should implement support for such patterns in the ucanto so it's possible to delegate all current and future capabilities.

Gozala:
Fixes #87

We need this to make [authorization](https://github.com/web3-storage/specs/blob/main/w3-session.md) protocol able to support [more granular delegations](https://hackmd.io/ZKvnIebsT3GWRvyNgH_pgw?view#Response) without specifying exact space DID. And we do not want to specify exact space DID to avoid having to recreate UCAN delegations chains and instead add desired proofs. Specifically consider following scenario

1. `did:key:zSpace` delegates to `did:mailto:web.mail:alice`
2. `did:mailto:web:mail:alice` authorizes `did:key:zAgent` 
3. `did:key:zAnotherSpace` delegates to `did:mailto:web.mail:alice`

Unless in step 2 `with` field is open ended like `did:*`, agent `did:key:zAgent` will not have access to new `did:key:zAnotherSpace` even though it would be able to get it through [access/claim](https://github.com/web3-storage/specs/blob/main/w3-access.md#claim-access).

Changes here implement native support for wildcards in `can` and `with` fields and consequently address above described limitation. Furthermore it eliminates need for [boilerplate capabilities we had to use](https://github.com/web3-storage/w3protocol/blob/99cbd2f9bbdb55f31773c5d654d02b66c2468e32/packages/capabilities/src/store.js#L15-L42)

Please note that implementation [correctly handles unintended escalation](https://github.com/web3-storage/ucanto/issues/87#issuecomment-1238752108) like:

```
{ can: "store/*", with: "did:*", size: 100 }
                ^
                |
{ can: "store/*", with: "did:*" }
                ^
                |
{ can:"store/add", with:"did:key:zAlice", size:200 }
```

To make this work when validator parses (middle) capability it inherits information from the outer (bottom) capability and comes out just like bottom one in this instance. By doing this validator correctly identifies escalation (200 > 100) in the next step.


packages/validator/test/capability-access.spec.js
```diff
@@ -0,0 +1,556 @@
...
+      can: 'dev/ping',
+      with: DID,
+      nb: {
+        message: Schema.string(),
+      },
+    }),
+  },
+}
+
+test('validates with patters', async () => {
```
gobengo, line 48:
```suggestion
test('validates with patterns', async () => {
```

is this typo?

packages/validator/test/capability-access.spec.js
```diff
@@ -0,0 +1,556 @@
...
+  const result = await access(await ping.delegate(), {
+    authority: w3,
+    capability: capabilities.dev.ping,
+    principal: Verifier,
+  })
+
+  assert.equal(result.error, undefined)
+})
+
+test('validates with patters in chain', async () => {
```
gobengo, line 79:
```suggestion
test('validates with patterns in chain', async () => {
```

packages/validator/test/capability-access.spec.js
```diff
@@ -0,0 +1,556 @@
...
+    issuer: jordan,
+    audience: w3,
+    with: alice.did(),
+    nb: {
+      message: 'hello',
+    },
+    proofs: [proof],
+  })
+
+  const result = await access(await ping.delegate(), {
```
gobengo, line 280: see https://github.com/web3-storage/ucanto/pull/218/files#r1100550414

IMHO it's not obvious to me from ucan-wg/spec that `middle` `dev/*` should authorize this, because spec doesn't say anything about `dev/*` syntax (nor `did:*`)

packages/validator/src/capability2.js
```diff
@@ -0,0 +1,109 @@
+import * as API from '@ucanto/interface'
```
gobengo, line 1: Is this `capability2.js` file imported anywhere?

gobengo: what is intended difference between `capability.js` and `capability2.js` (if any)

packages/validator/src/capability.js
```diff
@@ -635,29 +635,52 @@ class AndMatch {
...
+ * @param {T} match
+ * @param {string} pattern
+ * @returns {T|null}
+ */
+const matchAbility = (match, pattern) =>
+  match === pattern
+    ? match
+    : pattern === '*'
+    ? match
+    : pattern.endsWith('/*') && match.startsWith(pattern.slice(0, -1))
```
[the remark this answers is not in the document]
Gozala: > If I'm reading this right it seems like this will make it so foo/* always derives foo/bar by default.
>
> Is that part of the [ucan spec](https://github.com/ucan-wg/spec)? In there I only see special processing rules for can=* and can=ucan/* but not anything about any sort general {namespace}/* validation logic.

Correct UCAN spec prescribes no logic no how capabilities must be compared against one another and leaves it up application domain.

> I worry that adding that behavior into ucanto could, at worst, lead a programmer to accidentally authorize things in ucanto they intended to be unauthorized according to ucan spec.

I think this is a moot point, ucanto comes with other various opinions regarding UCANs, e.g. it considers any delegation for the resource matching DID of the issuer to be valid. The UCAN spec does not imply this, yet ucanto chooses to codify it.

Another example is special handling of `./update` capability. UCAN spec does not recognize such capability to be special in any way, yet ucanto validator chooses to recognize it and treat it in a special way.

I consider treatment of `ns/*` just the same, or more specifically ucanto validator would consider such delegations valid without having to manually thread logic in all the capability definitions. That said ucanto can be used with a different validator or no validator at all.

> I worry that adding that behavior into ucanto could, at worst, lead a programmer to accidentally authorize things in ucanto they intended to be unauthorized according to ucan spec.

I think we defend against this with more documentation, reviews tests etc... Given that UCAN spec does not define at all how capabilities are compared the risk of misusing validator is always there. 

Here we just codify how we already were thinking about capabilities, as a matter of fact there several cases where bugs were deployed because we assumed that wildcard patters worked this way while they were not e.g. see https://github.com/web3-storage/ucanto/issues/87#issuecomment-1275225435

> so e.g. I kinda liked how previously you had to explicitly derive in your parsers from {namespace}/* and e.g. it was possible/default to make a {namespace}/foo that was not authorized by {namespace}/*.

That only covered the bottom two in the below image, the chain below would fail to validate when it reached the top one, which is the problem because that is kind of arrangement you'll have with access protocol as described in the pr summary.

```
{ can: "store/*", with: "did:*", size: 100 }
                ^
                |
{ can: "store/*", with: "did:*" }
                ^
                |
{ can:"store/add", with:"did:key:zAlice", size:200 }
```

> @Gozala do you intend to specify that /*-suffix auto-derives logic in [ucan-wg](https://github.com/ucan-wg/spec) ?

Given that UCAN spec intentionally does not define any logic on how capabilities out to be compared it would make much sense IMO. Perhaps years from now there will be desire to have bit more structure around this, in that case I might attempt such a thing.

packages/validator/src/capability2.js
Gozala: oops, I did not intend to include that file

packages/validator/src/lib.js
```diff
@@ -30,14 +30,14 @@ export { Schema }
 /**
  * @param {UCAN.Link} proof
  */
-const unavailable = proof => new UnavailableProof(proof)
+export const unavailable = proof => new UnavailableProof(proof)
```
Gozala, line 33:
```suggestion
const unavailable = proof => new UnavailableProof(proof)
```

packages/validator/src/lib.js
```diff
@@ -30,14 +30,14 @@ export { Schema }
...
 
 /**
  *
  * @param {UCAN.DID} did
  * @returns {API.DIDKeyResolutionError}
  */
-const failDIDKeyResolution = did => new DIDKeyResolutionError(did)
+export const failDIDKeyResolution = did => new DIDKeyResolutionError(did)
```
Gozala, line 40:
```suggestion
const failDIDKeyResolution = did => new DIDKeyResolutionError(did)
```

packages/validator/src/lib.js
```diff
@@ -68,7 +68,7 @@ const resolveMatch = async (match, config) => {
  * @param {API.Proof[]} proofs
  * @param {Required<API.ProofResolver>} config
  */
-const resolveProofs = async (proofs, config) => {
+export const resolveProofs = async (proofs, config) => {
```
Gozala, line 71:
```suggestion
const resolveProofs = async (proofs, config) => {
```

packages/validator/src/lib.js
```diff
@@ -148,7 +148,7 @@ const resolveSources = async ({ delegation }, config) => {
  * @param {API.ParsedCapability} capability
  * @param {API.DID} issuer
  */
-const isSelfIssued = (capability, issuer) => capability.with === issuer
+export const isSelfIssued = (capability, issuer) => capability.with === issuer
```
Gozala, line 151:
```suggestion
const isSelfIssued = (capability, issuer) => capability.with === issuer
```

packages/validator/src/lib.js
```diff
@@ -447,7 +447,7 @@ class Unauthorized extends Failure {
  * @param {Required<API.ClaimOptions>} config
  * @returns {Promise<API.Result<T, API.InvalidProof|API.DIDKeyResolutionError>>}
  */
-const validate = async (delegation, config) => {
+export const validate = async (delegation, config) => {
```
Gozala, line 450:
```suggestion
const validate = async (delegation, config) => {
```

packages/validator/src/capability.js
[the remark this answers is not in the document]
gobengo: > That said ucanto can be used with a different validator or no validator at all.

good point

packages/validator/test/capability-access.spec.js
Gozala: So delegation chain we have here is as follows (where top on is ambient authority by ownership, which is also something UCAN spec has no opinion on)

```
{ with: did:key:zAli, can: * }
         ^
         :   - by Ownership to Alice
{ with: did:*, can: * }            =>  { with: did:key:zAli, can: * }
         ^
         |   - by Alice to Bob
{ with: did:*, can: dev/* }        => { with: did:key:zAli, can: dev/* }
         ^
         |   - by Bob to Mallory
{ with: did:*, can: * }            => { with: did:key:zAli, can: dev/* }
         ^
         |   - by Mallory to W3
{ with: did:key:zAli, can: 'dev/ping', nb: { message: 'hello'}  }
```

If you follow the chain you can see all the constraints that had been imposed at every layer (spec calls it diminishing) yet despite all the the constraints bottom is still true.

Now one could argue that Bob is escalating when delegating to Mallory, yet I think it goes against intuition and there is a discussion to spec this pattern because otherwise we have keep recreating chains every time some agent gains more capabilities, for details see:
https://github.com/ucan-wg/spec/issues/130
https://github.com/ucan-wg/spec/issues/131

gobengo (approved):

packages/validator/src/capability.js
```diff
@@ -635,29 +635,52 @@ class AndMatch {
 }
 
 /**
- * Parses capability `source` using a provided capability `parser`. By default
- * invocation parsing occurs, which respects a capability schema, failing if
- * any non-optional field is missing. If `optional` argument is `true` it will
- * parse capability as delegation, in this case all `nb` fields are considered
- * optional.
+ * @template {string} T
```
Gozala, line 638:
```suggestion
 * Takes an ability (can field of the capability) and an ability (wildcard) pattern
 * and returns back ability if pattern matches or `null` if it does not otherwise. 
 * @example
 * ```js
 * matchAbility('store/add', 'store/list') // => null
 * matchAbility('store/add', 'store/*') // => 'store/add'
 * matchAbility('store/add', '*') // => 'store/add'
 * matchAbility('store/add', 'store/') // => null
 * matchAbility('store/add', 'store/a*') // => null
 * ```
 * @template {API.Ability} T
```

packages/validator/src/capability.js
```diff
@@ -635,29 +635,52 @@ class AndMatch {
...
+  match === pattern
+    ? match
+    : pattern === '*'
+    ? match
+    : pattern.endsWith('/*') && match.startsWith(pattern.slice(0, -1))
+    ? match
+    : null
+
+/**
+ * @template {string} T
```
Gozala, line 653:
```suggestion
 * Function is used by the delegated ability parser to resolve `with` resource of the delegated
 * capability. It takes `match` argument corresponding to the `with` of the claimed capability
 * and a `pattern` corresponding to the `with` of the delegated capability. If pattern matches
 * returns `match` back, otherwise returns `null`.
 * @template {string} T
```

Gozala:
I have chatted with @expede about ways to express "everything, everywhere delegated or owned" resources we've decided we can express this as `ucan:*`. We'll likely spec `ucan:` protocol in more detail, but right now I just updated implementation replacing generic `URI` pattern with only `ucan:*` form. We can add support for other forms once they get speced.

alanshaw (commented):
Yeah this is awesome 🎉

PR #220, opened 2023-02-09
feat!: Use schema stuff in the capabilities instead of custom parsing

Gozala:
`capability` logic came before the `Schema` did we have not got around to integrating schema fully into `capability` we even end up with bunch of code that duplicated logic that is in schema.

Consequently it became impossible to use Schema.struct or Schema.dict on the `nb` field. This PR updates `capability` to just use `Schema`.

### DIFF
--- a/packages/validator/src/capability.js
+++ b/packages/validator/src/capability.js
@@ -225,6 +225,8 @@ class Capability extends Unit {
 }
 
 /**
+ * Normalizes capability by removing empty nb field.
+ *
  * @template {API.ParsedCapability} T
  * @param {T} source
  */
--- a/packages/validator/src/capability.js
+++ b/packages/validator/src/capability.js
@@ -7,7 +7,7 @@ import {
   DelegationError as MatchError,
   Failure,
 } from './error.js'
-import { invoke, delegate, DID } from '@ucanto/core'
+import { invoke, delegate } from '@ucanto/core'
 import * as Schema from './schema.js'
 
 /**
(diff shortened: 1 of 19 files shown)


packages/validator/src/capability.js
```diff
@@ -8,15 +8,37 @@ import {
...
  */
-export const capability = descriptor => new Capability(descriptor)
+export const capability = ({
+  derives = defaultDerives,
+  nb = defaultNBSchema,
+  ...etc
+}) => new Capability({ derives, nb, ...etc })
+
+const defaultNBSchema =
+  /** @type {Schema.MapRepresentation<any>} */
```
gobengo, line 40:
```suggestion
  /** @type {Schema.MapRepresentation<unknown>} */
```

not sure if this makes sense but just an idea..

packages/validator/test/capability.spec.js
```diff
@@ -1503,10 +1506,10 @@ test('capability with optional caveats', async () => {
   const Echo = capability({
     can: 'test/echo',
     with: URI.match({ protocol: 'did:' }),
-    nb: {
+    nb: Schema.struct({
```
gobengo, line 1509: wdyt of trying to make things backward incompatible so code like `nb: { message: URI }` still works even without wrapping with `Schema.struct`?

gobengo (approved):

packages/validator/test/capability.spec.js
Gozala: I wanted to, but amount of added complexity coupled with the fact that TS inference gets confused on occasions I don’t think it would be worth it

packages/validator/src/capability.js
Gozala: No this is not going to work I'm afraid because it's used as default parameter for generic object schema and anything but `any` would fail to meet that requirement.

Ideally TS would respect default type parameter and allow matching it with default parameter in the function, but sadly it doesn't work so this is the only workaround I'm aware of

PR #227, opened 2023-02-14
feat!: update session API

### LINKED ISSUE #226: Update `./update` capability to support extended fields
At the moment `./update` associates signing key with the principal. However we would like instead to check whether delegation is compliant with what's in the `./update` field by verifying that:

1. `aud` matches `nb.aud` field of the `./update` proof.
2. `att` match `nb.att` fields of the `./update` proof.

Gozala:
Fixes #226

## Overview

Changes in this PR update our session logic which previously just assigned a `did:key` to a `did:mailto` principal in the validation context.

Instead now we have `ucan/attest` capability that can be used to attest to the fact that specific delegation (that has non `did:key` issuer) has been authorized by the user. In other words when we perform interactive authorization flow like with `access/authorize` we can delegate `ucan/attest`  to the agent been authorized that proofs that we performed an out-of-bound authorization flow in which user has authorized delegation we're attesting to.

~~Now instead of assigning a key we set `authorization` CID, that corresponds to the UCAN Delegation without `proofs` and `signature`. With this changes when validator encounters delegation issued by principal other than did:key e.g  did:mailto it will attempt to find corresponding `./update` capability proof that sets `permit` to a CID derived from delegation. If such (valid) delegation is found, validator will not check signature and continue with UCAN verification process. If (session) delegation is found, validator falls back to doing key resolution instead.~~

~~All in all this gives us a way to authorize agent with a specific set of capabilities as opposed to giving it a sudo capabilities as it used to be the case previously.~~

~~P.S: I had to update Link schema so that validator could look for `./update` with a very specific CID in it as opposed to finding a mismatching one, which could be problem if there were multiple `./update` proofs, because validator stops as soon as it finds success path.~~

~~Link to the updated [spec](https://github.com/web3-storage/specs/pull/49)~~


packages/core/src/delegation.js
```diff
@@ -222,6 +259,161 @@ export const delegate = async (
...
+  facts = [],
+  nonce,
+}) => {
+  const value = {
+    iss: toPrincipal(issuer.did()),
+    aud: toPrincipal(audience.did()),
+    att: capabilities,
+    exp: expiration === Infinity ? null : expiration,
+    ...(facts.length > 0 && { fct: facts }),
+    // need to omit optionals or CBOR will throw
```
vasco-santos, line 294:
```suggestion
    // need to omit optionals or CBOR will throw
    ...(facts.length > 0 && { fct: facts }),
```

vasco-santos (approved):
LGTM! Great work

packages/interface/src/capability.ts
```diff
@@ -436,6 +436,12 @@ export interface InvalidSignature extends Failure {
   readonly delegation: Delegation
 }
 
+export interface SessionEscalation extends Failure {
```
gobengo, line 439: can you include a comment or link to doc on what 'escalation' means in this context? e.g. if I am debugging an error and see its this one, I'd want to read that to figure out how to interpret the error and understand if there is a chance I can do something appropriate if I catch it

gobengo: after some reading, I think I grok that it's a pretty generic error and if someone encounters it, retrying the request wont help because there's something wrong in their authorization

Gozala: Yes it's error that occurs when invalid `ucan/attest` is found. Invocation with same proof chain would indeed produce same results.

packages/validator/src/lib.js
```diff
@@ -464,87 +502,94 @@ const validate = async (delegation, config) => {
...
-    return result.error ? result : config.principal.parse(result).withDID(did)
-  }
+const verifySignature = async (delegation, verifier) => {
+  const valid = await UCAN.verifySignature(delegation.data, verifier)
+  return valid ? delegation : new InvalidSignature(delegation, verifier)
 }
 
 /**
- * @param {API.DID} did
+ * Attempts to find an authorization session - an `./update` capability
```
gobengo, line 569: does this `./update` still make sense?

can we link to which spec defines `ucan/attest` behavior?

packages/principal/src/account.js
```diff
@@ -0,0 +1,41 @@
...
+export const from = ({ id }) => new Account(id)
+
+/**
+ * An account is a special type of signer that produces empty signature, which
+ * signals that verifier needs to verify authorization interactively.
+ *
+ * @template {UCAN.DID} ID
+ * @implements {UCAN.Signer<ID, Signature.NON_STANDARD>}
+ */
+class Account {
```
gobengo, line 18: nit: if possible I'd avoid using `Account` at all here (it's too easy to misinterpret) and call this something like `NonStandardSigner`. just a thought

packages/principal/test/account.spec.js
```diff
@@ -0,0 +1,17 @@
+import { Account } from '../src/lib.js'
+import { assert } from 'chai'
+
+export const utf8 = new TextEncoder()
+describe('Account', async () => {
+  it('it can sign', async () => {
+    const account = Account.from({ id: 'did:mailto:alice@web.mail' })
```
gobengo, line 7: nit: should this test using the `did:mailto` syntax that disallows unescaped `@`?

packages/validator/src/lib.js
```diff
@@ -65,80 +65,111 @@ const resolveMatch = async (match, config) => {
 }
 
 /**
```
gobengo, line 67: all the comments in this file are awesome!! tysm

packages/validator/src/schema/link.js
```diff
@@ -24,7 +29,7 @@ class LinkSchema extends Schema.API {
    * @param {Settings<Code, Alg, Version>} settings
    * @returns {Schema.ReadResult<API.Link<unknown, Code, Alg, Version>>}
    */
-  readWith(cid, { code, algorithm, version }) {
+  readWith(cid, { code, multihash = {}, version }) {
```
gobengo, line 32: <3

packages/validator/test/session.spec.js
```diff
@@ -0,0 +1,312 @@
...
+        with: account.did(),
+        nb: {
+          message: 'hello world',
+        },
+      },
+    },
+  })
+})
+
+test('delegated ./update', async () => {
```
gobengo, line 78:
```suggestion
test('delegated ucan/attest', async () => {
```

gobengo (approved):

packages/principal/test/account.spec.js
Gozala, line 7:
```suggestion
    const account = Account.from({ id: 'did:mailto:web.mail:alice' })
```

packages/principal/test/account.spec.js
```diff
@@ -0,0 +1,17 @@
...
+    assert.deepEqual(account.did(), 'did:mailto:alice@web.mail')
```
Gozala, line 8:
```suggestion
    assert.deepEqual(account.did(), 'did:mailto:web.mail:alice')
```

packages/validator/src/lib.js
```diff
@@ -464,87 +502,94 @@ const validate = async (delegation, config) => {
...
+const verifySignature = async (delegation, verifier) => {
+  const valid = await UCAN.verifySignature(delegation.data, verifier)
+  return valid ? delegation : new InvalidSignature(delegation, verifier)
 }
 
 /**
- * @param {API.DID} did
+ * Attempts to find an authorization session - an `./update` capability
+ * delegation where `with` matches `config.authority` and `nb.permit`
+ * matches give delegation.
```
Gozala, line 571:
```suggestion
 * Attempts to find an authorization session - an `ucan/attest` capability
 * delegation where `with` matches `config.authority` and `nb.proof`
 * matches given delegation.
 * @see https://github.com/web3-storage/specs/blob/feat/auth+account/w3-session.md#authorization-session
```

packages/principal/src/account.js
Gozala: Renamed it to `Absentee` which I think reflect the better than either misleading `Account` or `NonStandardSigner` which isn't specific enough. I'll assume it's ok, if not we can rename in the followup change.

alanshaw (commented):
LGTM 👍

PR #229, opened 2023-02-20
proposal: add a bag of string-keyed key/value pairs to InvocationContext

travis:
As part of https://github.com/web3-storage/w3protocol/issues/347 I'm looking into making location information from Cloudflare HTTP headers available to the invocation handler that generates space registration emails (ie, https://github.com/web3-storage/w3protocol/blob/main/packages/access-api/src/service/voucher-claim.js#L45) so that we can tell users where in the world a space registration request was initiated. This is a common pattern in web services and is intended to help the user make better choices about which verification requests to approve.

To do this, we need some way of making HTTP headers available to invocation handlers. I'd like to do this without tying the InvocationContext too tightly to HTTP, and the simplest thing I can think of is adding a bag of string-keyed key/value pairs named `metadata` to InvocationContext.

Very open to other approaches here - this is intended to be the starting point for a conversation.

One note worth bringing up for clarity - this opens the door to invocation handlers doing all sorts of things with information from the HTTP headers, some of which could conceivably be a violation of a user's privacy. It may be worth implementing some sort of allow-list mechanism rather than just passing all HTTP headers through as I'm doing in this PR. I'm not entirely sure where this configuration would live, so am looking for thoughts and feedback.

### DIFF
--- a/packages/interface/src/lib.ts
+++ b/packages/interface/src/lib.ts
@@ -288,7 +288,12 @@ export interface InvocationContext extends CanIssue {
   my?: (issuer: DID) => Capability[]
   resolve?: (proof: UCANLink) => Await<Result<Delegation, UnavailableProof>>
 
-  principal: PrincipalParser
+  principal: PrincipalParser,
+  /**
+   * A bag of key/value pairs. Can be used, for instance, to make HTTP headers available
+   * to invocation handlers.
+   */
+  metadata?: Record<string, any>
 }
 
 export type ResolveServiceMethod<
--- a/packages/server/src/server.js
+++ b/packages/server/src/server.js
@@ -68,7 +68,7 @@ class Server {
  */
 export const handle = async (server, request) => {
   const invocations = await server.decoder.decode(request)
-  const result = await execute(invocations, server)
+  const result = await execute(invocations, server, request)
   return server.encoder.encode(result)
 }
 
@@ -78,16 +78,17 @@ export const handle = async (server, request) => {
  * @template {API.Tuple<API.ServiceInvocation<C, Service>>} I
  * @param {API.InferInvocations<I>} invocations
  * @param {API.ServerView<Service>} server
+ * @param {API.HTTPRequest} request
  * @returns {Promise<API.InferServiceInvocations<I, Service>>}
  */
-export const execute = async (invocations, server) => {
+export const execute = async (invocations, server, request) => {
   const results = []
   const input =
     /** @type {API.InferInvocation<API.ServiceInvocation<C, Service>>[]} */ (
       invocations
     )
   for (const invocation of input) {
-    results.push(await invoke(invocation, server))
+    results.push(await invoke(invocation, server, request))
   }
 
   return /** @type {API.InferServiceInvocations<I, Service>} */ (results)
@@ -98,9 +99,10 @@ export const execute = async (invocations, server) => {
  * @template {API.Capability} C
  * @param {API.InferInvocation<API.ServiceInvocation<C, Service>>} invocation
  * @param {API.ServerView<Service>} server
+ * @param {API.HTTPRequest} request
  * @returns {Promise<API.InferServiceInvocationReturn<C, Service>>}
  */
-export const invoke = async (invocation, server) => {
+export const invoke = async (invocation, server, request) => {
   // If invocation is not for our server respond with error
   if (invocation.audience.did() !== server.id.did()) {
     return /** @type {API.Result<any, API.InvalidAudience>} */ (
@@ -126,7 +128,7 @@ export const invoke = async (invocation, server) => {
     )
   } else {
     try {
-      return await handler[method](invocation, server.context)
+      return await handler[method](invocation, {...server.context, metadata: request.headers})
     } catch (error) {
       const err = new HandlerExecutionError(
         capability,


alanshaw:
@travis the ucanto server is created for each request so you _could_ just pass the headers as an additional property to the object passed to `Server.create` and it'll become the "server context" which is passed to your handler as the second argument.

i.e. add property when server is created here:

https://github.com/web3-storage/w3protocol/blob/8fd5a108ddf254b3f3c5fbfdde3575b3875567c6/packages/access-api/src/routes/root.js#L11

e.g.

```js
  const server = Server.create({
    id: env.signer,
    encoder: CBOR,
    decoder: CAR,
    service: service(env),
    headers: request.headers // <--- here
  })
```

Might be worth adding _request_ context (this PR) as well...

gobengo:
I recommend using the technique @alanshaw described in w3protocol and not making this change in ucanto, which I think brings in some httpish stuff to a layer we should keep agonstic to that (e.g. so handlers are agnostic to whether invoked via http or p2p)

gobengo:
I also tried to find some context for *why* we want to put that http header information in those emails sent by that voucher/claim handler. Like... what if that handler wasn't invoked via http? It's hard for me to tell what motivated https://github.com/web3-storage/w3protocol/issues/347 or, based on some stuff in notion, why it is important to do anything with http headers specifically, vs things that can be signed in the `voucher/claim` delegations or invocation.

travis:
@bengo ah for sure - 

>  why it is important to do anything with http headers specifically, vs things that can be signed in the voucher/claim delegations or invocation.

I could definitely be missing a path to doing this with something in the delegations or invocations, but the idea here is that we want to add information about where the request _our service_ received came from. As far as I know (and please point me in the right direction if I'm wrong, would love to push this to a different layer) the delegations and invocations are created and signed by the client, and we can't just take their word on where they're from.

I believe the reason Hugo formulated this this way is that Cloudflare already does some work to give us easily usable visitor location information in HTTP request headers they set:

https://support.cloudflare.com/hc/en-us/articles/200168236-Configuring-IP-geolocation

We could certainly do this without Cloudflare (or in non-HTTP transports) but at some level we need to look at the IP address of the request we received, geolocate it, and pass that information along to the invocation handler so it can be included in the email we send to users.

LMK if that helps, or if I'm missing anything/can clarify. I'll also be around morning PT if you want to hop on a huddle!

travis:
@alanshaw ah sweet, yea if we can get away with doing this mostly in w3protocol that would be great. Couple thoughts in my head as I noodle on this:

> the ucanto server is created for each request

Is this the intended usage for the `ucanto` `Server` class? ie, as `ucanto` library authors are we expecting consumers of this API to create a new `Server` per-request? I'm fine either way (or, tbh, with us having no preference) but I'd probably design this API slightly differently for different intentions on our end.

> Might be worth adding request context (this PR) as well...

I think this would make a lot of sense. Do you have any thoughts on what this context would look like? The `metadata Record<string, any>` field is where I ended up along this line of thought - I think you're saying that the change in this PR makes sense in this case, is that right?

travis:
@bengo - fwiw

> handlers are agnostic to whether invoked via http or p2p

is exactly why I ended up with this design (ie, adding a `metadata Record<string, any>` to the invocation request rather than anything HTTP-specific like a `HTTPRequest`) - any p2p transport would likely have similar metadata, and we'd be able to get an IP address that we could assign a key name from any network transport I can think of.

I definitely share the intention of not tying this any tighter to HTTP than absolutely necessary - would love any thoughts on how we might tweak this datastructure or data type to better meet that goal.

gobengo:
> is exactly why I ended up with this design (ie, adding a metadata Record<string, any> to the invocation request rather than anything HTTP-specific like a HTTPRequest) 

I should have left that comment attached to the code, but here is what I was referring to: https://github.com/web3-storage/ucanto/pull/229/files#diff-b8c2d26ed3fd075596a838c69b36c537536ae6973cafffdbf481fec464c9b6cfR81

gobengo:
> so that we can tell users where in the world a space registration request was initiated. This is a common pattern in web services and is intended to help the user make better choices about which verification requests to approve.

this also gives my email provider an unencrypted log of my location every time i log into any web app ...
I think we should apply privacy by design here and not store and forward this information unless we really really need to.

Personally, I don't think we should prioritize logging location compared to other candidates in https://github.com/web3-storage/w3protocol/issues/221 .

travis:
> I should have left that comment attached to the code, but here is what I was referring to: https://github.com/web3-storage/ucanto/pull/229/files#diff-b8c2d26ed3fd075596a838c69b36c537536ae6973cafffdbf481fec464c9b6cfR81

ahhh, yes yep agree - I think if we go with this design it makes sense to push the `metadata` field either into the server context somehow or up to the top level and not pass the request down through these handlers

travis:
> > so that we can tell users where in the world a space registration request was initiated. This is a common pattern in web services and is intended to help the user make better choices about which verification requests to approve.
> 
> this also gives my email provider an unencrypted log of my location every time i log into any web app 

That's true and a good point! 

To be honest don't have super strong feelings about this - there's a tradeoff between safety (ie, giving the user more factors to determine whether they should click on "approve") and privacy here and I think it's very much in our court to decide where we stand on it. 

@heyjay44 I'm inclined to put the brakes on the "Location info" task in https://github.com/web3-storage/w3protocol/issues/347 unless you or anyone else on the team wants to advocate for it. @gobengo is absolutely right that implementing this would leak user location information to third party email providers and I think that's something worth taking some time to think through and design/implement carefully to minimize harm. Thoughts?

travis:
closing as it sounds like we're going to hold off on this for now - happy to reopen if we decide to move forward

PR #251, opened 2023-03-07
feat!: remove canIssue default from the server

Gozala:


packages/validator/test/session.spec.js
```diff
@@ -310,3 +310,47 @@ test('resolve key', async () => {
     },
   })
 })
+
+test('service can not delegate access to account', async () => {
+  const account = Absentee.from({ id: 'did:mailto:web.mail:alice' })
```
gobengo, line 315: the other test didn't use `Absentee` and it doesn't seem used beyond `.did()` (just noting things that might make it different from other test)

packages/validator/test/session.spec.js
```diff
@@ -310,3 +310,47 @@ test('resolve key', async () => {
...
+  // service should not be able to delegate access to account resource
+  const auth = await Delegation.delegate({
```
gobengo, line 317: I used ucanto/core `delegate` export, which I believe ends up calling this, but was small diff

packages/validator/test/session.spec.js
```diff
@@ -310,3 +310,92 @@ test('resolve key', async () => {
...
+    capabilities: [
+      {
+        with: account.did(),
+        can: 'debug/echo',
+      },
+    ],
+  })
+
+  const session = await Delegation.delegate({
+    issuer: service,
```
gobengo, line 329: how is `service` different than `w3` used as issuer above?

gobengo (approved):

PR #257, opened 2023-03-12
feat: configurable audience handlers

### LINKED ISSUE #256: alllow server to handle requests for arbitrary audience
At the moment server rejects requests unless it matches `aud`. In most cases it is desired behavior, however there are some instances where one might want to handle or route requests when they don't match. For example it would be cool to receive requests which may be for `did:mailto` `aud` and forward those onto corresponding email address.

I think we could easily do it if we move the check from the server into provider code.

Gozala:
Fixes #256

### DIFF
--- a/packages/server/src/handler.js
+++ b/packages/server/src/handler.js
@@ -1,5 +1,5 @@
 import * as API from './api.js'
-import { access } from '@ucanto/validator'
+import { access, Schema, Failure } from '@ucanto/validator'
 
 /**
  * @template {API.Ability} A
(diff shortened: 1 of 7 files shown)


packages/server/src/handler.js
```diff
@@ -11,13 +11,37 @@ import { access } from '@ucanto/validator'
...
+ * @template {API.Caveats} C
+ * @template {unknown} U
+ * @param {object} input
+ * @param {API.Reader<API.DID>} [input.audience]
+ * @param {API.CapabilityParser<API.Match<API.ParsedCapability<A, R, C>>>} input.capability
+ * @param {(input:API.ProviderInput<API.ParsedCapability<A, R, C>>) => API.Await<U>} input.handler
+ * @returns {API.ServiceMethod<API.Capability<A, R, C>, Exclude<U, {error:true}>, Exclude<U, Exclude<U, {error:true}>>>}
+ */
+
+export const provideAdvanced =
```
gobengo, line 29: Could there be a comment that would help me understand how to choose between `provide` and `provideAdvanced` e.g. by defining what 'advanced' means here?

packages/server/test/handler.spec.js
```diff
@@ -242,3 +254,76 @@ test('test access/claim provider', async () => {
   const result = await claim.execute(client)
   assert.deepEqual(result, [])
 })
+
+test('handle did:mailto audiences', async () => {
```
gobengo, line 258: btw it would be nice to have a Schema for `did:mailto` that matches with `did:mailto:${string}:${string}` because `DID.match({ protocol: 'mailto' })` does not satisfy `did:mailto:${string}:${string}`

gobengo (approved):

packages/server/src/handler.js
Gozala: Added some code comments to address this

packages/server/test/handler.spec.js
Gozala: I agree, but that is probably out of scope for ucanto

Issue #254, opened 2023-03-11
Delegation should have method for listing capabilities it delegates


Gozala:
### Problem

Given that we have support for `can: *`  and `with: ucan:*` it is no longer clear just by looking at the `.capabilities` what resources and actions are delegated by the specific delegation. In most cases `{ can: *, with: ucan:* }` would be re-delegating  capabilities from the liked `proof`s.

### Proposal

It would be a good idea to introduce new function into the `@ucanto/core` that allows querying capabilities provided, which could look something along the lines of:

```ts
interface Extension {
  allows(delegation: Delegation, match?: Allows): Allows
}

interface Allows {
  [key:Resource]: {
     [key:Ability]: Record<string, unknown>
  }
}
```

Note that it maps the structure we're considering in next version of UCANs, but it also provides more convenient way to identify resources (spaces). Optional `match` also seems like good way to check for specific capability without having to iterate over results and what not.

I would expect implementation to go down the delegation chain until it's able to loose all the `*` or until it reaches the bottom.

It would be nice to provide a way to consider linked proofs that aren't included, but it's probably out of scope here. 

Doing validation is also tempting but I think it should be out of scope here as well.

gobengo:
sounds super useful!

Gozala:
> I think this is done via #259 . @Gozala wdyt?

Almost unlike proposal #259 does not allow you to select via match, so I'll keep this open until it's implemented

PR #259, opened 2023-03-14
feat: allow api inferring capabilities

Gozala:
Implements subset of #254 that can be used to derive set of capabilities provided by one or more delegations.


packages/core/src/delegation.js
```diff
@@ -21,6 +21,71 @@ export const isLink =
  */
 export const isDelegation = proof => !Link.isLink(proof)
 
+/**
+ * Takes on or more delegations and returns all delegated capabilities in
```
gobengo, line 25:
```suggestion
 * Takes one or more delegations and returns all delegated capabilities in
```

packages/core/src/delegation.js
```diff
@@ -21,6 +21,71 @@ export const isLink =
...
+ * UCAN 0.10 format, expanding all the special forms like `with: ucan:*` and
+ * `can: *` to an explicit forms.
```
gobengo, line 27:
```suggestion
 * `can: *` to explicit forms.
```

packages/core/src/delegation.js
```diff
@@ -21,6 +21,71 @@ export const isLink =
...
+ *
+ * @param {API.Delegation} delegation
+ * @returns {Iterable<API.Capability>}
+ */
+const iterateCapabilities = function* ({ issuer, capabilities, proofs }) {
+  for (const { with: uri, can, nb } of capabilities) {
+    if (uri === 'ucan:*') {
+      yield {
+        with: issuer.did(),
+        can,
```
gobengo, line 68: if `can` is `*` here, it seems it would not be expanded

packages/core/src/delegation.js
```diff
@@ -21,6 +21,71 @@ export const isLink =
...
+        nb,
+      }
+
+      for (const proof of proofs) {
+        if (isDelegation(proof)) {
+          for (const capability of iterateCapabilities(proof)) {
+            yield {
+              with: capability.with,
+              can: can === '*' ? capability.can : can,
+              nb: { ...capability.nb, ...Object(nb) },
```
gobengo, line 78: It seems like the latter spread of `nb` should be more `can`-value-aware.
Does it really make sense to merge `Object(nb)` over `capability.nb` even if their `can` values are super different?

gobengo: I think I can see how, if it did lead to any surprising behavior, it looks like it would err on the side of over-attenuating, which I'm fine with compared to the alternative.

packages/core/src/delegation.js
```diff
@@ -21,6 +21,71 @@ export const isLink =
...
+      // delegated
+      const capabilities =
+        capability.with === 'ucan:*'
+          ? iterateCapabilities(delegation)
+          : [capability]
+
+      for (const { with: uri, can, nb } of capabilities) {
+        const resource = allow[uri] || (allow[uri] = {})
+        const abilities = resource[can] || (resource[can] = [])
+        abilities.push({ ...Object(nb) })
```
gobengo, line 48: Do we also have enough info here to include something like the relevant subset of `delegations` you'd need to exercise this capability? e.g. if so we could put a method on here that iterates them back out, in case that would be useful for putting only the proofs necessary on an invocation

packages/interface/src/lib.ts
```diff
@@ -177,6 +177,105 @@ export interface Delegation<C extends Capabilities = Capabilities> {
   delegate(): Await<Delegation<C>>
 }
 
+/**
+ * Type representing a UCAN capability set in UCAN 0.10 format.
+ * @see https://github.com/ucan-wg/spec/blob/0.10/README.md#241-examples
```
gobengo, line 182: great comments throughout these new types tysm

gobengo (approved):
couple questions above, but no need to block

packages/core/src/delegation.js
Gozala: Yes that is correct, however that is expected here as well. When you say `{ "with": "ucan:*", "can": "*" }` it means I re-delegate everything I own or have been delegated.

This specific yield is everything I own, which is `{ with: my.did(), can: "*" }`.

packages/core/src/delegation.js
Gozala: You are right actual merge logic depends on the capability schema, however at this layer we don't know what the schema is so we can do accurate merge.

Here we just do naive merge which will either be correct one or later will be rejected by (capability semantics aware) validator because `nb` escalates.

Which is I think the best option available given the constraints

packages/core/src/delegation.js
Gozala: Added code comment to clarify this

packages/core/src/delegation.js
Gozala: Added code comment to clarify this

travis:
this looks awesome! I've started using it in my local so I'm excited for this to ship. once question - you aren't re-exporting the new function from `lib.js` so the only way I could use this locally was like:

```js
import * as ucanto from '@ucanto/core'
...
export function canDelegateCapability(delegation, child) {
  const allowsCapabilities = ucanto.Delegation.allows(delegation)
...
```

which typescript still isn't super happy with (but it compiles and runs just fine): 

```
Property 'allows' does not exist on type 'typeof import("/Users/travis/dev/pl/w3protocol/node_modules/.pnpm/@ucanto+core@5.1.0/node_modules/@ucanto/core/dist/src/delegation")'.ts(2339)
```

could just be a transient TS failure but curious if you think this function is worth re-exporting manually like we do with `delegation` and `isDelegation` in https://github.com/web3-storage/ucanto/blob/feat/cap-query/packages/core/src/lib.js ?

Gozala:
> ```
> 
> ```
> 
> 
>     
>   
> 
> could just be a transient TS failure but curious if you think this function is worth re-exporting manually like we do with `delegation` and `isDelegation` in https://github.com/web3-storage/ucanto/blob/feat/cap-query/packages/core/src/lib.js ?

I suspect that error happens because typedefs have not been generated when you linked locally so it still refers to and old version.

In other words I would expect `ucanto.Delegation.allows` to work fine. For now I don't want to expose this using `ucanto.allows` as I'm not confident we're going to keep it as is, so making it less prominent seems like a better idea.

gobengo (approved):

PR #266, opened 2023-03-28
feat!: implement receipts

### LINKED ISSUE #265: Implement receipts at the transport layer
Doing full receipt support proved to be difficult as it requires many breaking changes. Instead we should be able to implement receipts only at the transport layer minimizing breaking changes elsewhere.

Gozala:
# Overview

- Implements Receipts as per [upcoming invocation spec 0.2](https://github.com/ucan-wg/invocation/blob/v0.2/README.md)
  - [ ] Few things changed since so I need to update some code
  - ⚠️ clients now return `Receipt` as opposed to `Result` type which is a breaking change.
- Changes transport codecs so that they can adapt based on content-type / accept headers
   - Now we have `Transport.Codec` as opposed to `{ encoder, decoder }` pair which performs content type negotiation.
   - ⚠️ Now client takes `codec` field corresponding to `Transport.OutboundCodec`
   - ⚠️ Now server takes` codec` field corresponding to `Transport.InboundCodec`
   - ⚠️ CBOR transport is gone on server we could use new `Legacy` codec to deal with all clients and new clients should not use CBOR.
- CAR and CBOR codecs moved to `@ucanto/core` as they are needed for receipts. Transport still reexports them so it's not breaking change. 
- `@ucanto/core` now re-exports things that were used from multiformats to cut down deps and avoid things getting out of sync.
- Handlers can no longer return falsy values, it was a best practice that now is ennfroced by the typesystem and the library. If falcy value is returned it will end up encoded as unit `{}` in the receipt.
- CARs didn't used to contains roots in blocks, now they do because previous version complicated encode / decode of receipts for no good reason
- `@ucanto/core` now has exports `DAG` module with bunch of utilities to for building and viewing DAGs encoded in CARs. 

Fixes #265


packages/client/src/connection.js
```diff
@@ -41,12 +40,44 @@ class Connection {
...
-  return result
+  // We may fail to decode the response if content type is not supported
+  // or if data was corrupted. We do not want to throw in such case however,
+  // because client will get an Error object as opposed to a receipt, to retain
+  // consistent client API with two kinds of errors we encode caught error as
+  // a receipts per workflow invocation.
+  try {
+    return await connection.codec.decode(response)
+  } catch (error) {
+    const { message, ...cause } = /** @type {Error} */ (error)
```
gobengo, line 57: I believe technically `error` could be undefined and the destructuring would lead to a `SyntaxError` in that case. (might be wise to check `typeof error === 'object'` first.
unlikely but a pain if/when it does.

packages/server/src/server.js
```diff
@@ -111,21 +116,41 @@ export const invoke = async (invocation, server) => {
...
+        // form at is unit type `{}`.
+        result: /** @type {API.ReceiptResult<{}>} */ (
+          value?.error ? { error: value } : { ok: value == null ? {} : value }
+        ),
+      })
+    } catch (cause) {
+      const error = new HandlerExecutionError(
         capability,
-        /** @type {Error} */ (error)
+        /** @type {Error} */ (cause)
```
gobengo, line 144: It feels brittle to type cast this here. There's no guarantee that `cause` matches `Error` afaict.

packages/server/test/handler.spec.js
```diff
@@ -107,14 +106,12 @@ test('checks service id', async () => {
...
-    decoder: CAR,
-    encoder: CBOR,
+    codec: CAR.inbound,
   })
 
   const client = Client.connect({
     id: w3,
-    encoder: CAR,
-    decoder: CBOR,
+    codec: CAR.outbound,
```
gobengo, line 114: It would be great if client could interrogate `options.channel` to pick a good default for this if possible. (e.g. if `options.channel` is an HTTP Channel, it could issue an HTTP request to the URL interrogating 'which codecs do you support' e.g. `options.channel.supportedCodecs(clientSupports)` to get back a supported codec intersection)

(we could ship CAR codec with client so that always works, and if someone passes more supported codecs, it works as long as the supported codecs intersection has at least 1)

gobengo: Not saying we should do now, but I hope one day we can make creating ucanto clients not require extra a priori knowledge about how which codecs are required.

packages/transport/src/car/request.js
```diff
@@ -0,0 +1,69 @@
+import * as API from '@ucanto/interface'
+import { CAR } from '@ucanto/core'
+import { Delegation } from '@ucanto/core'
+
+export { CAR as codec }
+
+const HEADERS = Object.freeze({
+  'content-type': 'application/car',
```
gobengo, line 8: seems like this media type should be defined in `@ipld/js-car` or otherwise stored in a constant somewhere

packages/transport/src/codec.js
```diff
@@ -0,0 +1,211 @@
...
+/**
+ * @implements {API.InboundCodec}
+ */
+class Inbound {
+  /**
+   * @param {API.HTTPRequest} request
+   * @returns {API.ReceiptResult<API.InboundAcceptCodec, API.HTTPError>} transport
+   */
+  accept({ headers }) {
+    const contentType = headers['content-type'] || headers['Content-Type']
```
gobengo, line 22: http headers are defined as case insensitive. We should probably only check for `content-type`. `cOnTent-TypE` is also as valid as `headers['Content-Type']` and it seems reasonable to expect normalization at a higher layer

gobengo (approved):

packages/interface/src/transport.ts
```diff
@@ -22,16 +23,20 @@ export interface EncodeOptions {
   readonly hasher?: UCAN.MultihashHasher
 }
 
+export interface RequestEncodeOptions extends EncodeOptions {
+  accept?: string
```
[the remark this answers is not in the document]
gobengo: (or e.g. if comma separated values are allowed too). comment or jsdoc example would illustrate

packages/server/test/handler.spec.js
gobengo: By reading further I think I realized that the `outbound` object is already set up to do something like that!
cool. I'd be down for `CAR.outbound` to just be a default value here

packages/server/src/handler.js
```diff
@@ -10,7 +10,7 @@ import { access, Schema, Failure } from '@ucanto/validator'
  * @template {API.Ability} A
  * @template {API.URI} R
  * @template {API.Caveats} C
- * @template {unknown} U
+ * @template {{}} U
```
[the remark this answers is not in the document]
Gozala: Yes my intention is to disallow null and undefined

packages/client/src/connection.js
Gozala:
```suggestion
    // No third party code is run during decode and we know
    // we only throw an Error
    const { message, ...cause } = /** @type {Error} */ (error)
```

packages/client/src/connection.js
```diff
@@ -41,12 +40,44 @@ class Connection {
...
+    const receipts = []
+    for await (const invocation of workflow) {
+      const { cid } = await invocation.delegate()
+      const receipt = await Receipt.issue({
```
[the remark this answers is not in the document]
Gozala: I agree this is not ideal, yet I feel very strongly that two channels for error handling is going to be a super error prone. 

```ts
try {
   const receipt = await agent.invoke(capbility)
    if (receipt.out.error) {
         // handle error
    }
} catch (error) {
   // handle error
}
```

Perhaps we could come up with a better solution so that they are separate steps, but lets do it in a followup

packages/client/test/client.spec.js
```diff
@@ -244,21 +262,64 @@ test('execute with delegations', async () => {
...
+test('decode error', async () => {
+  const client = Client.connect({
+    id: w3,
+    channel,
+    codec: Codec.outbound({
+      encoders: {
+        'application/car': CAR.request,
+      },
+      decoders: {
+        'application/car+receipt': CAR.response,
```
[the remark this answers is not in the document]
Gozala: hmm `application/car` had been used all over the place, but looks like IPFS gateway now sets `application/vnd.ipld.car;version=1` and so should we.

Complication however here is that our codecs also used to set this as content type before this change so not recognizing it will break existing clients. I created a followup issue for this https://github.com/web3-storage/ucanto/issues/269

packages/core/src/dag.js
```diff
@@ -0,0 +1,102 @@
...
+import * as MF from 'multiformats/interface'
+import * as CBOR from './cbor.js'
+
+/**
+ * @param {unknown} value
+ * @returns {IterableIterator<API.Block>}
+ */
+export const iterate = function* (value) {
+  if (
+    value &&
```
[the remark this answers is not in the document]
Gozala: hmm intention here was to pass anything and if it is an `IPLDView` it would iterate over it's blocks and if it isn't do nothing. It is to traverse arbitrary structures that may contain `IPLDView`s inside. We could instead have `isIPLDView(object)` predicate so it would be:

```ts
for (const [key, value] of Object.entries(something)) {
    if (isIPLDView(value)) {
        for (const block of value.iterateIPLDBlocks()) {
        }
    }
}
```

But I find it more convenient to do this instead

```ts
for (const [key, value] of Object.entries(something)) {
    for (const block of DAG.iterate(value)) {
    }
}
```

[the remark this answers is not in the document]
Gozala: Added a comment to make it bit more clear.

packages/core/src/invocation.js
```diff
@@ -8,6 +9,30 @@ import { delegate } from './delegation.js'
  */
 export const invoke = options => new IssuedInvocation(options)
 
+/**
```
Gozala, line 12:
```suggestion
/**
 * Takes a link of the `root` block and a map of blocks and constructs an
 * `Invocation` from it. If `root` is not included in the provided blocks it
 * throws an error. If root points to wrong block (that is not an invocation)
 * it will misbehave and likely throw some errors on field access.
 *
```

packages/core/src/receipt/outcome.js
```diff
@@ -0,0 +1,127 @@
...
+ * @returns {API.Outcome<Ok, Error, Ran>}
+ */
+export const view = ({ root, blocks }) => {
+  return new Outcome({
+    root: DAG.decodeFrom(root, blocks),
+    store: blocks,
+  })
+}
+
+/**
```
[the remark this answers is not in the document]
Gozala: 0.2 end up inlining this block into receipt and I have branch that remove this, so I won't bother commenting

packages/core/test/invocation.spec.js
```diff
@@ -154,5 +154,6 @@ test('execute invocation', async () => {
     },
   })
 
+  // @ts-expect-error
```
Gozala, line 157:
```suggestion
  // Note that types expect us to return receipt but here we
  // just test that execute simply returns `connection.execute(this)[0]` which
  // will not be a receipt due to our implementation above. 
  // @ts-expect-error
```

packages/interface/src/lib.ts
```diff
@@ -318,6 +377,136 @@ export type LinkJSON<T extends UnknownLink = UnknownLink> = ToJSON<
...
+ * invocation. It is issued and signed by the task executor or its delegate.
+ *
+ * @see https://github.com/ucan-wg/invocation#8-receipt
+ */
+export interface ReceiptModel<
+  Ok extends {} = {},
+  Error extends {} = {},
+  Ran extends Invocation = Invocation
+> {
+  ocm: Link<OutcomeModel<Ok, Error, Ran>>
```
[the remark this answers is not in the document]
Gozala: Comment above links to the spec it implements which lists out the `ocm` field. This structure represents decoded IPLD block so fields will be exactly as in the schema.

packages/core/src/dag.js
```diff
@@ -0,0 +1,108 @@
...
+import * as CBOR from './cbor.js'
+
+/**
+ * Function takes arbitrary value and if it happens to be an `IPLDView`
+ * it will iterate over it's blocks. It is just a convenience for traversing
+ * arbitrary structures that may contain `IPLDView`s in them. 
+ * Note if you pass anything other than `IPLDView` it will not attempt
+ * to find views nested inside them, instead it will just emit no blocks.
+ * @param {unknown} value.
+ *
```
Gozala, line 14:
```suggestion
```

PR #267, opened 2023-03-28
feat: allow additional authorities to verify invocation

alanshaw:
This seems to work good for me.

You might want to thread this deeper, but let me know if you want to merge as is and I'll write a test to fix the coverage.


Gozala:
> You might want to thread this deeper

I wrote up a issue describing what I think would be an ideal solution well aligned with some other features in the pipeline.

> but let me know if you want to merge as is and I'll write a test to fix the coverage.

That said I don't mind taking this as, I just would like to mark added `authorities` option `@deprecated` to signal that it will be replaced in the future when implementation per #268 is done.

PR #271, opened 2023-03-30
feat: align implementation with receipt 0.2 spec

Gozala:
Main change that happened since last iteration is that `Outcome` used be a separate block and now it is just a field on `Receipt` block https://github.com/ucan-wg/invocation/blob/v0.2/README.md#82-receipt 

Less obvious change that happened is addition of `ReceiptBulider` which is basically a placeholder structure so you could defer issuing a receipt. Currently it is only used by `issue` function, but it is set up such that you could allow passing different codecs and hashers. I hope to do something similar for `Delegation` as well in the future so that you could send a request with `accept: application/json` and ucanto will respond in JSON as opposed to cbors in the the car.


packages/core/test/receipt.spec.js
```diff
@@ -85,7 +84,7 @@ test('receipt with proofs', async () => {
     ],
   })
 
-  const { cid: proofCid } = await delegate({
+  const { cid: proofCid } = await await delegate({
```
Gozala, line 87:
```suggestion
  const { cid: proofCid } = await delegate({
```

packages/core/src/receipt.js
```diff
@@ -1,79 +1,122 @@
...
+  const data = CBOR.decode(bytes)
 
-  return new Receipt({ root: block, store: blocks, outcome })
+  return new Receipt({ root: { bytes, cid, data }, store: blocks })
 }
 
 /**
+ * Represents a UCAN invocation receipt view over some block store e.g. in
+ * memory CAR. It incrementally decodes proofs, ran invocation etc. on access
+ * which reduces overhead but potentially defers errors if references blocks
```
gobengo, line 28:
```suggestion
 * which reduces overhead but potentially defers errors if referenced blocks
```

gobengo (approved):

PR #272, opened 2023-04-01
feat!: update ucanto to invocation spec compatible result type

Gozala:
## Overview

- Updates result type to use ucan invocation spec compatible representation
- At the library layer (except of the schemas which needs to deal with optional fields) we now enforce no `null|undefined` values in Results and push towards use of unit value `{}` instead.
   - Mostly this guides users into a better and more extensible API design.
   - Please note that `false` and `""` are still allowed.
- ⚠️ above changes caused breaking API change on `.derives` methods than no longer return `true` or `Failure` but instead return `Result<{}, Failure>`.
- ⚠️ Error types no longer have `error: true` field which were in place to pattern match on ok / error.


gobengo (approved):

gobengo:
This is a big change to use the `{ ok }` or `{ error }` result types explicitly, but I think it's for the best in the long run to use these explicit types.

packages/core/src/result.js
```diff
@@ -0,0 +1,51 @@
+import * as API from '@ucanto/interface'
+
+/**
```
Gozala, line 3:
```suggestion
/**
 * Creates the success result containing given `value`. Throws if
 * `null` or `undefined` passed to encourage use of units instead.
 *
```

packages/core/src/result.js
```diff
@@ -0,0 +1,51 @@
...
+  if (cause == null) {
+    throw new TypeError(
+      `error(${cause}) is not allowed, consider passing an error instead`
+    )
+  } else {
+    return { error: cause }
+  }
+}
+
+/**
```
Gozala, line 31:
```suggestion
/**
 * Creates the failing result containing an error with a given
 * `message`. Unlike `error` function it creates a very generic
 *  error with `message` & `stack` fields. The `error` function
 * is recommended over `fail` for all but the most basic use cases.
 *
```

packages/interface/src/capability.ts
```diff
@@ -54,11 +54,7 @@ export interface MatchSelector<M extends Match>
 
 export interface DirectMatch<T> extends Match<T, DirectMatch<T>> {}
 
-export interface Reader<
-  O = unknown,
-  I = unknown,
-  X extends { error: true } = Failure
-> {
+export interface Reader<O = unknown, I = unknown, X extends {} = Failure> {
```
Gozala, line 57:
```suggestion
/**
 * Generic reader interface that can be used to read `O` value form the
 * input `I` value. Reader may fail and error is denoted by `X` type.
 *
 * @template O - The output type of this reader
 * @template I - The input type of this reader.
 * @template X - The error type denotes failure reader may produce.
 */
export interface Reader<O = unknown, I = unknown, X extends {} = Failure> {
```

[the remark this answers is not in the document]
Gozala: Added jsdoc comments with details

PR #273, opened 2023-04-07
feat!: migrate schema to core

Gozala:
# Overview

- Migrates schema from validator to the core
   - I needed it in the transport layer to deal with workflow versions etc.. and pulling in validator there made no sense.
   - Creating separate PR only with this to reduce amount of changes bundled with transport logic
- Adds Variant schema for working with Variant types


packages/core/test/schema/fixtures.js
```diff
@@ -1,6 +1,7 @@
 import { pass, fail, display } from './util.js'
 import * as Schema from '../../src/schema.js'
-import { string, unknown } from '../../src/schema.js'
+
+const { string, unknown } = Schema
```
Gozala, line 4:
```suggestion

import { string, unknown } from '../../src/schema.js'
```

packages/core/test/schema/fixtures.js
```diff
@@ -1,5 +1,6 @@
 import { pass, fail, display } from './util.js'
 import * as Schema from '../../src/schema.js'
+
```
Gozala, line 3:
```suggestion
```

packages/core/src/schema/type.ts
```diff
@@ -103,6 +103,83 @@ export interface StructSchema<
   partial(): MapRepresentation<Partial<InferStruct<U>>, I> & StructSchema
 }
 
+/**
+ * Schema for the {@link Variant} types, which is a special kind of union type
+ * where every choice is a struct with a single field denoting the choice.
+ *
+ * The `_` branch is a special case which can be used to represent all other
```
gobengo, line 110: Is there a reason we have to have the `_` special case, instead of specifying this default behavior as a whole other parameter not inside `VariantChoices`?
One downside is it seems to preclude using `_` as a variant name (not a huge loss, but seems worth avoiding if possible). Is there an upside to specifying the default behavior via a reserved key name?

gobengo (approved):

packages/core/src/schema/type.ts
Gozala: It is very common in typed languages with pattern matching to use `_` as a "default" branch e.g. [rust does this](https://doc.rust-lang.org/book/ch18-03-pattern-syntax.html), [elm does this](https://exercism.org/tracks/elm/concepts/pattern-matching) haskell does this, even ucan invocation spec does it in some sense, so it seemed reasonable to do it here.

> One downside is it seems to preclude using _ as a variant name (not a huge loss, but seems worth avoiding if possible).

I think using `_` as branch name is bad enough idea that I don't actually feel it's a real loss.

 > specifying this default behavior as a whole other parameter not inside `VariantChoices`

It is possible and I have tried that in fact, but it made inference types and `variant` function a lot more complicated and brittle because you ended up with this unnamed branch. In the end it felt clunky so I've chose to go with `_` instead.

gobengo: > but it made inference types and variant function a lot more complicated and brittle

I was guessing maybe this was why (but wanted to confirm), and +1 for avoiding the brittleness.
I didn't realize that `_` is common in other languages. Thanks for explaining that.

PR #274, opened 2023-04-07
feat!: versioned transport protocol

Gozala:
## Overview

- Changes introduce `AgentMessage` block which are DAGs send around between ucanto (client/server) agents.
- `AgentMessage`  may contain set of delegations (to be invoked) and / or set of receipts.
   - At the moment clients send invocations and servers respond with receipts but in the future we may allow sending both in either direction.
- `AgentMessage` is represented as a variant case so we could introduce another version in the future (e.g. when ucan invocations are implemented)
- Transport encoders / decoders now simply encode / decode `AgentMessage`-es and no longer know or care what they contain.
- New transport codecs now set proper CAR content type while legacy codec uses application/car.
   - This end up been useful way to differentiate between legacy requests that put invocation links in roots vs new requests that put message cid into a root.


packages/server/test/server.spec.js
```diff
@@ -417,3 +418,53 @@ test('falsy errors are turned into {}', async () => {
     ok: {},
   })
 })
+
+test('run invocation without encode / decode', async () => {
```
gobengo, line 422: awesome!

gobengo (approved):
`AgentMessage` is a good abstraction over the various ways ucanto can be configured to send messages (e.g. so apps using ucanto can transition from `legacy` to ucan invocation spec )

PR #287, opened 2023-04-18
feat: archive/extract api for delegations

Gozala:
For the workshop I have end up implementing these methods so we could archive ucan chains into bytes and extract them back. This pr adds them to the into the ucanto

### DIFF
--- a/packages/core/src/delegation.js
+++ b/packages/core/src/delegation.js
@@ -216,6 +216,9 @@ export class Delegation {
     return exportDAG(this.root, this.blocks, this.attachedLinks)
   }
 
+  /**
+   * @returns {API.Await<API.Result<Uint8Array, Error>>}
+   */
   archive() {
     return archive(this)
   }
@@ -317,6 +320,7 @@ export class Delegation {
  * buffer and returns it.
  *
  * @param {API.Delegation} delegation}
+ * @returns {Promise<API.Result<Uint8Array, Error>>}
  */
 export const archive = async delegation => {
   try {
@@ -416,7 +420,7 @@ const decode = ({ bytes }) => {
  */
 
 export const delegate = async (
-  { issuer, audience, proofs = [], attachedBlocks = new Map, ...input },
+  { issuer, audience, proofs = [], attachedBlocks = new Map(), ...input },
   options
 ) => {
   const links = []
--- a/packages/core/src/delegation.js
+++ b/packages/core/src/delegation.js
@@ -370,7 +370,7 @@ export const extract = async archive => {
     if (root == null) {
       return Schema.error('CAR archive does not contain a root block')
     }
-    const { bytes } = DAG.get(root.cid, blocks)
+    const { bytes } = root
     const variant = CBOR.decode(bytes)
     const [, link] = ArchiveSchema.match(variant)
     return ok(view({ root: link, blocks }))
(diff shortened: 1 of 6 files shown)


packages/core/src/delegation.js
```diff
@@ -288,6 +296,69 @@ export class Delegation {
...
+    // Then we we create a descriptor block to describe what this DAG represents
+    // and it to the block store as well.
+    const variant = await CBOR.write({
+      [`ucan@${delegation.version}`]: delegation.root.cid,
+    })
+    store.set(`${variant.cid}`, variant)
+
+    // And finally we encode the whole thing into a CAR.
+    const bytes = CAR.encode({
+      roots: [variant],
```
alanshaw, line 323: This takes a block not a CID?

packages/core/src/delegation.js
```diff
@@ -288,6 +296,69 @@ export class Delegation {
...
+    for (const block of delegation.iterateIPLDBlocks()) {
+      store.set(`${block.cid}`, block)
+    }
+
+    // Then we we create a descriptor block to describe what this DAG represents
+    // and it to the block store as well.
+    const variant = await CBOR.write({
+      [`ucan@${delegation.version}`]: delegation.root.cid,
+    })
+    store.set(`${variant.cid}`, variant)
```
alanshaw, line 319: Seems like in `CAR.encode` the root blocks are included (i.e. I think you might get the root block twice in the encoded CAR).

packages/core/src/delegation.js
```diff
@@ -288,6 +296,69 @@ export class Delegation {
...
+    })
+
+    return ok(bytes)
+  } catch (cause) {
+    return error(/** @type {Error} */ (cause))
+  }
+}
+
+export const ArchiveSchema = Schema.variant({
+  'ucan@0.9.1': /** @type {Schema.Schema<API.UCANLink>} */ (
```
alanshaw, line 334: Should this version number be a constant somewhere?

packages/core/src/delegation.js
```diff
@@ -288,6 +296,69 @@ export class Delegation {
...
+ * @param {Uint8Array} archive
+ */
+export const extract = async archive => {
+  try {
+    const { roots, blocks } = CAR.decode(archive)
+    const [root] = roots
+    if (root == null) {
+      return Schema.error('CAR archive does not contain a root block')
+    }
+    const { bytes } = DAG.get(root.cid, blocks)
```
alanshaw, line 353: This is already a block - decode retrieves it.

```suggestion
    const { bytes } = root
```

alanshaw (changes requested):

packages/core/src/delegation.js
Gozala: It is the `UCAN.VERSION`, however at some point we'll probably upgrade it to next version, yet version here should remain the same. I'm not sure there is a value in defining a new constant somewhere if this is the only thing referring to it though.

packages/core/src/delegation.js
Gozala: No you won't and I plan on changing CAR codec so that roots are no longer blocks but rather CIDs. Current approach of roots been blocks are kind of falling apart if you get a car from someone that has a root that isn't inside the car

PR #288, opened 2023-04-24
feat: support inline blocks in invocation

vasco-santos:
This PR adds support to create and encode invocations adding inline blocks to the encoded bytes for transport. This enables us to add the bytes for a IPLD Block with a CAR file over the wire for projects like w3 aggregation

Example of usage delegation
```js
const ucan = await Delegation.delegate({
    issuer: alice,
    audience: bob,
    capabilities: [
      {
        can: 'store/add',
        with: alice.did(),
      },
    ],
  })

const block = await getBlock({ test: 'inlineBlock' })
ucan.attach(block)
```

 invocation:

```js
const add = invoke({
    issuer: alice,
    audience: w3,
    capability: {
      can: 'store/add',
      with: alice.did(),
      link: 'bafy...stuff',
    },
    proofs: [],
  })

const block = await getBlock({ test: 'inlineBlock' })
add.attach(block)
```

### DIFF
--- a/packages/core/src/delegation.js
+++ b/packages/core/src/delegation.js
@@ -161,8 +161,9 @@ export class Delegation {
   constructor(root, blocks = new Map()) {
     this.root = root
     this.blocks = blocks
-    /** @type {API.BlockStore<unknown>} */
-    this.attachedBlocks = new Map()
+
+    /** @type {API.AttachedLinkSet} */
+    this.attachedLinks = new Set()
 
     Object.defineProperties(this, {
       blocks: {
@@ -203,14 +204,15 @@ export class Delegation {
    * @param {API.Block} block
    */
   attach(block) {
-    this.attachedBlocks.set(`${block.cid}`, block)
+    this.attachedLinks.add(`${block.cid}`)
+    this.blocks.set(`${block.cid}`, block)
   }
   export() {
-    return exportDAG(this.root, this.blocks)
+    return exportDAG(this.root, this.blocks, this.attachedLinks)
   }
 
   iterateIPLDBlocks() {
-    return exportBlocks(this.root, this.blocks, this.attachedBlocks)
+    return exportDAG(this.root, this.blocks, this.attachedLinks)
   }
 
   /**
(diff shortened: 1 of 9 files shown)


Gozala (changes requested):
I might be missing something but I find adding blocks to the ucanto `message` awkward as clients and servers operate at the layer above it. I am also not sure that adding `inlineLinks` to the message interface is a good idea, instead I would encourage surfacing links in invoked capabilities.

I would like to propose alternative design that:

1. Adds `attach(block: Block)` method to the `Delegation` / `Invocation` so that one could attach additional blocks to be included with the invocation.
2. Update `iterateIPLDBlocks` implementation on the `Delegation` class to emit "attached" blocks so they will be included with a `AgentMessage` and a CAR file.

I think that would meet all of the requirements and pave the way to the future where ucanto can figure out which blocks to attach from the schema itself.

vasco-santos:
Thanks @Gozala 

This is super useful direction. I agree that makes more sense to add this out of the message scope. Just refactored based on review

packages/core/src/delegation.js
```diff
@@ -192,12 +194,20 @@ export class Delegation {
     Object.defineProperties(this, { data: { value: data, enumerable: false } })
     return data
   }
+  /**
+   * Attach a block to the invocation encoded bytes.
```
Gozala, line 198:
```suggestion
   * Attach a block to the invocation DAG so it would be included in the
   * block iterator.
   * ⚠️ You should only attach blocks that are referenced from the `capabilities`
   * or `facts`, if that is not the case you probably should reconsider.
```

Nit: I think this is more clear

Gozala (approved):
I made couple of suggestions & would leave it up to you to decide whether to address them or keep things as is.

My only concern is that once invocation goes over the wire and is de-serialized on the other side it will loose attachments because `replica.iterateIPLDBlocks` will not be aware of the attachments.

I think it would be a good idea to capture these links in the data model somewhere, which leaves us with two options:

1. In the `iterateIPLDBlocks` traverse `capabilities` (or perhaps `this.data` and include all linked blocks).
2. Have dedicated place for sticking attached block links so we can retain behavior of `iterateIPLDBlocks` after they cross the wire. For example we could stick `{ 'ucanto/attachments': Link[] }` object into `fct` fields and use it instead of traversing arbitrary dags.

I don't think we need to block this on the ☝️ but we decide to leave that out for now lets create an issue for it and maybe a code comment in `attach` function to call out the limitation.

vasco-santos:
@Gozala addressed inline feedback, added comment for attach limitation and created https://github.com/web3-storage/ucanto/issues/296 for follow up. Going to get this now merged

Issue #296, opened 2023-04-27
Capture attachments in the data model to support de-serialized instance with blocks


vasco-santos:
> My only concern is that once invocation goes over the wire and is de-serialized on the other side it will loose attachments because replica.iterateIPLDBlocks will not be aware of the attachments.

> I think it would be a good idea to capture these links in the data model somewhere, which leaves us with two options:

> In the iterateIPLDBlocks traverse capabilities (or perhaps this.data and include all linked blocks).
Have dedicated place for sticking attached block links so we can retain behavior of iterateIPLDBlocks after they cross the wire. For example we could stick { 'ucanto/attachments': Link[] } object into fct fields and use it instead of traversing arbitrary dags.

Context: https://github.com/web3-storage/ucanto/pull/288#pullrequestreview-1402307369

vasco-santos:
@Gozala integrating this now into a client and server and we should get this iteration too.

```js
aggregate: {
  offer: Server.provide(AggregateCapabilities.offer, ({ invocation }) => {
    // get block by CID needed
  }
}
```

I also wonder if we should support a convenience API to just read block by CID, thoughts?

Regarding two pointed out solutions, I would favour a non magical approach, therefore second seems more explicit. We could mutate facts adding attachments when `.attach()` is called. However, only invocation has them. How will we do with delegations? What do you think?

Also one question, could you give me a high level hint on how the deserialization flow is until getting a invocation object created for the server.provide function?

Gozala:
Just to summarize to what we have arrived at during our call about this:

We could amend `exportDAG` functionality so that instead of iterating `attachedLinks` field it would iterate over links in the `capabilities` and `facts` fields. In fact `attachedLinks` can probably be turned into a lazy getter that collects all those links into a set and then `exportDAG` will remain as is. If we do that I would also like if `attach` method did not allow you adding blocks that aren't in the `attachedLinks`.

https://github.com/web3-storage/ucanto/blob/f72e6540a2fc323e40fe8078b9216215485e0397/packages/core/src/delegation.js#L401-L410


We do not need a separate API to dereference attached blocks because:

1. We will have `attachedLinks` getter that will tell you all the links.
2. We already have `blocks` field that is a map of blocks so you could de-reference from there.


https://github.com/web3-storage/ucanto/blob/f72e6540a2fc323e40fe8078b9216215485e0397/packages/interface/src/lib.ts#L210-L220

PR #298, opened 2023-04-28
fix: capture attachments links from capabilities and facts

vasco-santos:
Captures attachments links from capabilities and facts, so that we can de-serialize invocations and get the attached blocks again if they were attached. Based on discussion https://github.com/web3-storage/ucanto/issues/296#issuecomment-1526072090

Also note that:
- attach method does not allow you adding blocks that aren't in the attachedLinks.
- we now do not care in `exportDag` if blocks were not attached when exporting. Once we move next step forward and make this part of the schema, this should throw if block is required by schema

### DIFF
--- a/packages/core/src/delegation.js
+++ b/packages/core/src/delegation.js
@@ -162,14 +162,44 @@ export class Delegation {
     this.root = root
     this.blocks = blocks
 
-    /** @type {API.AttachedLinkSet} */
-    this.attachedLinks = new Set()
-
     Object.defineProperties(this, {
       blocks: {
         enumerable: false,
       },
     })
+  }
+
+  /**
+   * @returns {API.AttachedLinkSet}
+   */
+  get attachedLinks() {
+    const _attachedLinks = new Set()
+    const ucanView = decode(this.root)
+
+    // Get links from capabilities nb
+    for (const capability of ucanView.capabilities) {
+      /** @type {Link[]} */
+      const links = Object.values(capability.nb || {})
+        .filter(e => Link.isLink(e))
+
+      for (const link of links) {
+        _attachedLinks.add(`${link}`)
+      }
+    }
+
+    // Get links from facts values
+    for (const fact of ucanView.facts) {
+      /** @type {Link[]} */
+      // @ts-expect-error isLink does not infer value type
+      const links = Object.values(fact)
+        .filter(e => Link.isLink(e))
+
+      for (const link of links) {
+        _attachedLinks.add(`${link}`)
+      }
+    }
+
+    return _attachedLinks
   }
 
   get version() {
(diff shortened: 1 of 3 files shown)


packages/core/src/delegation.js
```diff
@@ -162,16 +162,46 @@ export class Delegation {
...
       },
     })
   }
 
+  /**
+   * @returns {API.AttachedLinkSet}
+   */
+  get attachedLinks() {
+    const _attachedLinks = new Set()
+    const ucanView = decode(this.root)
```
Gozala, line 177:
```suggestion
    const ucanView = this.data
```

we already have a getter for this

packages/core/src/delegation.js
```diff
@@ -162,16 +162,46 @@ export class Delegation {
...
+      for (const link of links) {
+        _attachedLinks.add(`${link}`)
+      }
+    }
+
+    // Get links from facts values
+    for (const fact of ucanView.facts) {
+      /** @type {Link[]} */
+      // @ts-expect-error isLink does not infer value type
+      const links = Object.values(fact)
```
Gozala, line 194: I would also check `isLink(fact)`

packages/core/src/delegation.js
```diff
@@ -162,16 +162,46 @@ export class Delegation {
...
+   * @returns {API.AttachedLinkSet}
+   */
+  get attachedLinks() {
+    const _attachedLinks = new Set()
+    const ucanView = decode(this.root)
+
+    // Get links from capabilities nb
+    for (const capability of ucanView.capabilities) {
+      /** @type {Link[]} */
+      const links = Object.values(capability.nb || {})
```
Gozala, line 182: I think it might be a good idea to do a full traversal to capture nested links e.g. because we have couple of instances where that is the case

packages/core/src/delegation.js
```diff
@@ -198,14 +228,15 @@ export class Delegation {
...
    * block iterator.
-   * ⚠️ You should only attach blocks that are referenced from the `capabilities`
-   * or `facts`, if that is not the case you probably should reconsider.
-   * ⚠️ Once a delegation is de-serialized the attached blocks will not be re-attached.
+   * ⚠️ You can only attach blocks that are referenced from the `capabilities`
+   * or `facts`.
    *
    * @param {API.Block} block
    */
   attach(block) {
```
Gozala, line 236: Can we modify this to take `IPLDView` instead ? That way we can attach things that span multiple blocks and allow us to use this for `access/delagate` and `access/claim` apis

Gozala (approved):
Looks good to me, I did suggest some improvements though

packages/core/src/delegation.js
vasco-santos: I will add this as new PR given it is a breaking change.

PR #300, opened 2023-04-28
feat: add Schema.bytes

Gozala:
In the IPFS Thing workshop we needed schema for bytes which I have injected into observable. This change ports it into ucanto.


vasco-santos (approved):
LGTM

PR #303, opened 2023-05-02
fix!: attach ipld view instead of single block

vasco-santos:
Attach view instead of single block per https://github.com/web3-storage/ucanto/pull/298#discussion_r1181285696

BREAKING CHANGE: delegation.attach() and invocation.attach() now receive IPLDView instead of Block

### DIFF
--- a/packages/core/src/delegation.js
+++ b/packages/core/src/delegation.js
@@ -228,18 +228,20 @@ export class Delegation {
     return data
   }
   /**
-   * Attach a block to the delegation DAG so it would be included in the
-   * block iterator.
+   * Attach blocks from IPLD view to the delegation DAG so it would be included
+   * in the block iterator.
    * ⚠️ You can only attach blocks that are referenced from the `capabilities`
    * or `facts`.
    *
-   * @param {API.Block} block
+   * @param {API.IPLDView} view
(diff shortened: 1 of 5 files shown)


Gozala (changes requested):
Thanks for tackling this. Unfortunately I'm no longer sure this is a good idea and maybe leaving it at the block level now is a better option. This is why

- Right now capabilities & facts are not fully traversed implying that after transporting the delegation (transitive) nodes, linked from the attachment will be lost, unless they were explicitly linked from the `capability.nb` or `facts`.
   - Although `attach` right now throws  if transitive blocks aren't linked, but then it's not clear to me if there is any benefit to just allowing only a block.

What I was picturing when suggesting this was that capabilities and facts would only link to the attachment roots. That however would require deep traversal of the `capabilities` and `facts` as opposed to shallow one implemented now.

In other words I think we should either do full traversal, otherwise this going to be more error prone than helpful. However I could be missing something so maybe we can talk sync about this.

vasco-santos:
@Gozala we are now doing full traversal of capabilities. Also checking facts as an entry in array, or value of an object. I guess, we only then need to do a full traversal of facts?

PR #310, opened 2023-05-22
feat!: update `@ucanto/interface` dependencies

travis:
`@ucanto/interface` had breaking changes, but because code in dependant libraries didn't change, release-please didn't cut new releases.

by manually specifying that the `@ucanto/interface` dependency must by 8 or above, this should trigger a release-please flow to release `@ucanto/interface@8`-compatible releases of these four libs.

### DIFF
--- a/packages/principal/package.json
+++ b/packages/principal/package.json
@@ -29,7 +29,7 @@
   "dependencies": {
     "@ipld/dag-ucan": "^3.3.2",
     "@noble/ed25519": "^1.7.3",
-    "@ucanto/interface": "workspace:^",
+    "@ucanto/interface": "workspace:^8",
     "multiformats": "^11.0.0",
     "one-webcrypto": "^1.0.3"
   },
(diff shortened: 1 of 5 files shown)


packages/principal/package.json
```diff
@@ -29,7 +29,7 @@
   "dependencies": {
     "@ipld/dag-ucan": "^3.3.2",
     "@noble/ed25519": "^1.7.3",
-    "@ucanto/interface": "workspace:^",
+    "@ucanto/interface": "workspace:^8",
```
Gozala, line 32: I don't think anything here was affected so not sure we need to force update

Gozala (approved):
Looks good to me. I don't think principal got affected and needs bump but it's fine either way. I think we probably want to revert this after things get released.

packages/principal/package.json
travis: I think you're right - I'm going to just go ahead and release it to make the upgrade a little easier for downstream folks (ie, "update everything to 8" rather than "update almost everything to 8")

travis:
> I think we probably want to revert this after things get released.

I think that's correct - will follow up with a `chore` PR once this is released

PR #317, opened 2023-08-03
fix: add missing dependency on `@ucanto/principal`

alanshaw:
Imported by `server.js`

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+auto-install-peers=false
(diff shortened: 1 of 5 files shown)


.npmrc
```diff
@@ -0,0 +1 @@
+auto-install-peers=false
```
alanshaw, line 1: see https://github.com/pnpm/pnpm/issues/6649

vasco-santos (approved):

PR #318, opened 2023-09-01
feat: implement support for BLS keys

Gozala:
Implement bls principal so it could be used for filecoin deal signing


packages/principal/src/bls/signer.js
```diff
@@ -0,0 +1,243 @@
...
+ * @param {string} archive
+ */
+export const fromFilecoinWallet = archive => {
+  const bytes = base16.baseDecode(archive)
+  const data = JSON.parse(new TextDecoder().decode(bytes))
+  if (data.Type !== 'bls') {
+    throw new TypeError(`Unsupported key type ${data.Type}`)
+  }
+
+  // For whatever reason bytes order is reversed
```
vasco-santos, line 120: Is there any point in a spec or whatever related to this?

packages/principal/src/bls/type.ts
```diff
@@ -0,0 +1,76 @@
...
+ */
+export type SigAlg = MulticodecCode<typeof Signature.BLS12381G2, 'BLS12381G2'>
+
+/**
+ * Name corresponding to BLS12381G2 algorithm.
+ */
+export type Name = 'BLS12381G2'
+
+/**
+ * This interface parametrizes {@link SignerKey} and extends it with Ed specific
```
vasco-santos, line 25: with Ed specific looks copy paste and not needed?

packages/principal/src/bls/type.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export interface BLSSignature<T = unknown> extends SignatureView<T, SigAlg> {}
+
+/**
+ * This interface parametrizes {@link VerifierKey} and extends it with BLS
+ * specific details.
+ */
+export interface BLSVerifier extends VerifierKey<SigAlg> {
+  /**
+   * Multicodec code that corresponds to Ed public key.
```
vasco-santos, line 68:
```suggestion
   * Multicodec code that corresponds to BLS public key.
```

packages/principal/src/bls/type.ts
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * This interface parametrizes {@link SignerKey} and extends it with Ed specific
+ * details.
+ */
+export interface BLSSigner extends SignerKey<SigAlg> {
+  readonly signatureAlgorithm: Name
+  readonly signatureCode: SigAlg
+  /**
+   * Multicodec code that corresponds to Ed private key.
```
vasco-santos, line 32:
```suggestion
   * Multicodec code that corresponds to BLS private key.
```

vasco-santos (approved):
LGTM

packages/principal/src/bls/signer.js
Gozala: I have not found anything and spend hours trying to figure out why things were incompatible until I have came across this https://github.com/drand/tlock-js/blob/410e80691e236f25318749122a7ab6e6f6e2d8dc/src/crypto/utils.ts#L37-L38

My guess is that's how go serializes them, but I have not managed to find actual code that does it to check that assumbtion.

packages/principal/src/bls/type.ts
Gozala:
```suggestion
 * This interface parametrizes {@link SignerKey} and extends it with BLS specific
```

Issue #331, opened 2023-11-17
when loading @ucanto/core through skypack.dev "TypeError: The specifier “@ucanto/interface” was a bare specifier, but was not remapped to anything"


gobengo:
https://observablehq.com/d/de1d4c6cac193f23

It worked in 5.x https://observablehq.com/@gobengo/w3up-ucanto

gobengo:
I thinkt hat back in 5.x, the src/lib.js for core didn't try to export * from `@ucanto/interface`. https://github.com/web3-storage/ucanto/blob/core-v5.2.0/packages/core/src/lib.js

PR #332, opened 2023-11-17
fix: update src/lib.js of @ucanto/interface to have an export statement

gobengo:
could fix https://github.com/web3-storage/ucanto/issues/331
> TypeError: The specifier “@ucanto/interface” was a bare specifier, but was not remapped to anything. Relative module specifiers must start with “./”, “../” or “/”.

Rationale
* `@ucanto/interface` package.json has `main` set to `src/lib.js` but that file is empty. and has no exports, which sometimes can confuse js-sniffing tools trying to figure out whether its a module or commonjs or what


gobengo:
happens here through jsdelivr https://jsfiddle.net/

gobengo:
I don't think this will help it necessarily

gobengo:
No, this does seem both safe and worth trying so the file contents are obviously a js module.

gobengo:
https://github.com/web3-storage/ucanto/issues/331#issuecomment-1816844350

PR #339, opened 2024-01-31
feat: add Receipt#toJSON

gobengo:
Motivation:
* start conversation about whether this is a good idea
* rationale:
  * ucanto Receipts have many getters that are 'views' into other objects (e.g. they might lazily decode on demand). Invocations/Delegations can be like this too.
  * when these objects get JSON stringified, the JSON representation only shows the underlying data, not the lazily decoded properties from getters. But frequently I am logging the object in order to see the derived/lazy properties, not just the underlying one.
* workaround: https://github.com/web3-storage/migrate-to-w3up/pull/3/files#r1473527173

### DIFF
--- a/packages/core/src/receipt.js
+++ b/packages/core/src/receipt.js
@@ -28,6 +28,25 @@ export const view = ({ root, blocks }, fallback) => {
 }
 
 /**
+ * @template {{}} Ok
+ * @template {{}} Error
+ * @template {API.Invocation} Ran
+ * @template {API.SigAlg} [SigAlg=API.SigAlg]
+ * @param {Receipt<Ok, Error, Ran, SigAlg>} r
+ */
+function toJSON(r) {
+  return {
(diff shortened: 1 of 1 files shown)


packages/core/src/receipt.js
```diff
@@ -27,6 +27,25 @@ export const view = ({ root, blocks }, fallback) => {
   return new Receipt({ root: { ...block, data }, store: blocks })
 }
 
+/**
+ * @template {{}} Ok
+ * @template {{}} Error
+ * @template {API.Invocation} Ran
+ * @template {API.SigAlg} [SigAlg=API.SigAlg]
+ * @param {Receipt<Ok, Error, Ran, SigAlg>} r
```
gobengo, line 35:
```suggestion
 * @param {Receipt<Ok, Error, Ran, SigAlg>} r
 * @returns {unknown}
```

PR #341, opened 2024-02-21
fix: revert "feat: http transport includes more information in error message when encountering non-ok http response"

gobengo:
Reverts web3-storage/ucanto#340


gobengo:
Reverting this because I think I merged in such a way that the commit message in main was not conventional enough for release-please to work https://github.com/web3-storage/ucanto/actions/runs/7995883699/job/21837196577#step:2:110

I plan to un-revert as a followup to this.

PR #347, opened 2024-04-03
feat!: add support for embedded effects

Gozala:
Adds changes to support embedding effects into receipts. Behavior is similar to the way `proofs` work in the delegations. If you pass an invocation it will be embedded in a agent message, if you pass link to invocation no invocation will be embedded.


vasco-santos (approved):
lgtm ✨  thanks for putting this together

vasco-santos:
I am merging and releasing it to integrate on my PRs

PR #360, opened 2025-01-17
feat(core): add schema type for reading did bytes

hannahhoward:
In regular UCAN fields, DIDs are often represented as bytes, and we have libraries for encoding and parsing them as bytes in @ipld/dag-ucan/did

However, the schema parser only supports the string form, so that is what is used in capabilities. It is a confusing difference.

This adds a bytes form of DIDs for schema parsing.

Perhaps @Gozala has context on why the deicision was originally made this way. 

Also, before this gets merged, I think we should revisit the right decision to make here, in particular before we start publishing content on the indexer -- maybe we should just stick to strings and fix the go code now using bytes.


alanshaw:
In the encoded UCAN, the issuer and audience DIDs are both encoded as bytes.

See:
https://github.com/ipld/js-dag-ucan/blob/21bae4e0ad7ff76e0c7b73938dcc228726854a43/src/schema.js#L235-L236
https://github.com/ipld/js-dag-ucan/blob/21bae4e0ad7ff76e0c7b73938dcc228726854a43/src/schema.js#L40-L41

AFAIK `Schema.did()` expects strings because the resource (`with`) in UCAN <=0.9 is meant to be a `URI`. We always use a DID (in UCAN 1.0 it _has_ to be a DID) so we had to encode as string here to comply with the spec.

I couldn't find an instance where we (Storacha) actually use a DID in caveats. The location commitment is the first. I believe that's why we don't have a thing in `Schema` for it yet.

IMHO it's reasonable (and consistent) to use the byte form when a DID appears in caveats, albeit less readable when inspecting a dag-json decoded UCAN for debugging.

packages/core/src/schema.js
```diff
@@ -4,6 +4,6 @@ export * as DID from './schema/did.js'
 export * as Text from './schema/text.js'
 export * from './schema/schema.js'
 export { match as link } from './schema/link.js'
-export { match as did } from './schema/did.js'
+export { match as did, matchBytes as didBytes} from './schema/did.js'
```
alanshaw, line 7:
```suggestion
export { match as did, matchBytes as didBytes } from './schema/did.js'
```

alanshaw (approved):
LGTM

PR #369, opened 2025-02-13
feat: validate attestation from another service

### LINKED ISSUE #268: Support for multiple authorities with distinct DIDs
Creating an issue that https://github.com/web3-storage/ucanto/pull/267 is attempting to fix.

@alanshaw use case is to accept attestations from a DID other than one passed as `authority` to a server / validator. However there is also generally a need to recognize and verify signatures from non `did:key:` principals. I think we could address both cases if we do following:

1. We already have an option that can be used to resolve `did:key` principal from other (non `did:key`)

    https://github.com/web3-storage/ucanto/blob/86332db62d00ca72a989da5640a35a844d111ab9/packages/interface/src/lib.ts#L650-L659
    https://github.com/web3-storage/ucanto/blob/86332db62d00ca72a989da5640a35a844d111ab9/packages/validator/src/lib.js#L544-L557

    These hook could be threaded through the server allowing us to support injecting keys for multiple services e.g. `web3.storage` and `nft.storage`. However these hooks alone would not be enough to accept attestations from the various actors for that we need no 2.

2. We need to be able to pass set of proofs to the server / validator that could authorize specific authorities. There are lot of use cases for that (e.g. worker running may need to have different key from the actual service key in which case service needs to delegate to the worker). To support attestations specifically we could utilize `proofs` and specifically pass delegations from authority to the principal trusted to do an attestation on it's behalf which would look like:

   ```ts
   Server.create({
      id: w3,
      proofs: [
         await delegate({
            issuer: w3,
            audience: notary,
            capabilities: [{
               can: "ucan/attest",
               with: w3.did()
            }]
         })
      ]
   })
   ```

   Assuming those are threaded through to the validator we could change `verifySession` function in the validator to recognize attestations from all the authorized principals and not just `authority`:
   
   https://github.com/web3-storage/ucanto/blob/86332db62d00ca72a989da5640a35a844d111ab9/packages/validator/src/lib.js#L581-L597

   In other words we would derive `with` schema as union of all the DIDs that authority has delegated to so something like
   
   ```ts
   const dids = config.proofs
      .filter(ucan => ucan.capabilities[0].with === config.authority.with && ucan.capabilities[0].can === 'ucan/attest')
      .map(ucan => Schema.literal(ucan.audience.did()))
   const withSchema = Schema.union(dids)
   ```

No 2 will ensure that attestations issued by authorized principals are recognized and respected. No 1 will ensure that signature chains could be verified which will be needed if attestation is issued by did principal that isn't did:key.

alanshaw:
This PR implements #268 

Specifically, we are changing the `did:web:web3.storage` service DID to `did:web:up.storacha.network` and want to allow attestations issued by `did:web:web3.storage` to still be considered valid.

resolves #268 
closes #267

### DIFF
--- a/packages/core/src/schema/schema.js
+++ b/packages/core/src/schema/schema.js
@@ -563,7 +563,7 @@ class Union extends API {
  * @param {U} variants
  * @returns {Schema.Schema<Schema.InferUnion<U>, I>}
  */
-const union = variants => new Union(variants)
+export const union = variants => new Union(variants)
 
 /**
  * @template T, U
--- a/packages/interface/src/capability.ts
+++ b/packages/interface/src/capability.ts
@@ -1,6 +1,7 @@
 import { Ability, Capability, DID, Link, Resource } from '@ipld/dag-ucan'
 import * as UCAN from '@ipld/dag-ucan'
 import {
+  AuthorityProver,
   Delegation,
   Result,
   Failure,
@@ -376,7 +377,8 @@ export interface ValidationOptions<
     PrincipalOptions,
     PrincipalResolver,
     ProofResolver,
-    RevocationChecker {
+    RevocationChecker,
+    Partial<AuthorityProver> {
   capability: CapabilityParser<Match<C, any>>
 }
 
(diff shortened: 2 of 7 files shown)


hannahhoward (approved):
This is a reasonable endeavor.

PR #371, opened 2025-02-13
feat: allow alternative audience for server

alanshaw:
This PR adds a new server option `audience`, which allows the server to accept invocations for audiences other than the service itself.

This option was already available (and continues to be available) on a per handler basis (via `provideAdvanced`), it takes precedence over the new global server audience.

Concretely, the DID for web3.storage is changing, but we want to be able to accept invocations targeting the old DID without having to configure every single handler.

### DIFF
--- a/packages/server/test/server.spec.js
+++ b/packages/server/test/server.spec.js
@@ -387,6 +387,56 @@ test('did:web principal resolve', async () => {
   })
 })
 
+test('alternative audience', async () => {
+  const car = await CAR.codec.write({
+    roots: [await CBOR.write({ hello: 'world ' })],
+  })
+  const alias = bob.withDID('did:web:alias.storage')
+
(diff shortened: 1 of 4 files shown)


packages/server/test/server.spec.js
```diff
@@ -387,6 +387,56 @@ test('did:web principal resolve', async () => {
...
+
+  const connection = Client.connect({
+    id: server.id,
+    codec: CAR.outbound,
+    channel: server,
+  })
+
+  const identify = Client.invoke({
+    issuer: alice,
+    audience: alias,
```
fforbeck, line 419: **Suggestion**: Add an invocation from `alias` to `alice` audience to have both branches of the audience schema validated.

fforbeck (approved):
Looks good to me!

PR #375, opened 2025-04-01
fix: Return 400 Bad Request for malformed payloads instead of handler execution error

### LINKED ISSUE #358: Malformed payload results in handler execution error, not bad request (400)
Decoding input payload [throws](https://github.com/storacha-network/ucanto/blob/6de27c0f19790d9def1df2f2e299fa4f0996ded9/packages/server/src/server.js#L82) and is [caught](https://github.com/storacha-network/ucanto/blob/6de27c0f19790d9def1df2f2e299fa4f0996ded9/packages/server/src/server.js#L153-L167) resulting in a handler execution error.

This could be returned as a HTTP response (similar to what happens when calling `accept`) with status 400 (bad request) to communicate the issue to users a bit better.

Crosstons:
Currently, when a request has a malformed payload, the decoder throws an exception which is caught and returned as a handler execution error. This change improves error handling by specifically catching decoder exceptions and returning a proper HTTP 400 Bad Request response with a meaningful error message, making it clearer to clients that the issue is with their request format. Closes #358

### DIFF
--- a/packages/server/src/server.js
+++ b/packages/server/src/server.js
@@ -79,10 +79,18 @@ export const handle = async (server, request) => {
     }
   } else {
     const { encoder, decoder } = selection.ok
-    const message = await decoder.decode(request)
-    const result = await execute(message, server)
-    const response = await encoder.encode(result)
-    return response
+    try {
+      const message = await decoder.decode(request)
+      const result = await execute(message, server)
+      const response = await encoder.encode(result)
+      return response
+    } catch (error) {
+      return {
+        status: 400,
+        headers: { 'Content-Type': 'text/plain' },
+        body: new TextEncoder().encode(`Bad request: Malformed payload - ${error.message || 'Unable to decode request'}`),
+      }
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


packages/server/src/server.js
```diff
@@ -79,10 +79,19 @@ export const handle = async (server, request) => {
     }
   } else {
     const { encoder, decoder } = selection.ok
-    const message = await decoder.decode(request)
-    const result = await execute(message, server)
-    const response = await encoder.encode(result)
-    return response
+    try {
+      const message = await decoder.decode(request)
```
alanshaw, line 83: We only need the try/catch around the decode, otherwise legitimate handler execution errors will also be returned as a 400.

alanshaw (changes requested):

alanshaw (changes requested):
Can you add a test for this to retain coverage?

Crosstons:
Hey @alanshaw ! I've added tests for the malformed payload error handling. The tests cover both cases
1. When the decoder throws an Error instance
2. When the decoder throws a non-Error value

Both tests verify that we get a proper 400 Bad Request response with a clear error message. This should help maintain the error handling behavior you wanted.

Let me know if you'd like me to make any adjustments to the tests!

PR #376, opened 2025-04-02
feat(transport): support custom headers in HTTP transport

fforbeck:
### Changes
Added the ability to set default headers when creating HTTP channels that get merged with request headers. 

Request headers take precedence over channel headers.
This will allow us to pass custom headers to track which tool the agents are running (CLI, console, etc).

See: https://github.com/storacha/ai-integrations/issues/48

### DIFF
--- a/packages/transport/src/http.js
+++ b/packages/transport/src/http.js
@@ -19,9 +19,10 @@ import * as API from '@ucanto/interface'
  * @param {URL} options.url
  * @param {(url:string, init:API.HTTPRequest) => API.Await<FetchResponse>} [options.fetch]
  * @param {string} [options.method]
+ * @param {Headers | Map<string, string> | Record<string, string>} [options.headers]
  * @returns {API.Channel<S>}
  */
-export const open = ({ url, method = 'POST', fetch }) => {
+export const open = ({ url, method = 'POST', fetch, headers }) => {
   /* c8 ignore next 9 */
   if (!fetch) {
     if (typeof globalThis.fetch !== 'undefined') {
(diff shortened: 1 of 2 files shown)


packages/transport/src/http.js
```diff
@@ -19,9 +19,10 @@ import * as API from '@ucanto/interface'
  * @param {URL} options.url
  * @param {(url:string, init:API.HTTPRequest) => API.Await<FetchResponse>} [options.fetch]
  * @param {string} [options.method]
+ * @param {Headers | Map<string, string> | Record<string, string>} [options.headers]
```
alanshaw, line 22: I would just use `Record<string, string>` as this is the existing type for the `headers` that can be passed to the `request()` method - i.e. as defined by `HTTPRequest`. This simplifies merging significantly.

packages/transport/src/http.js
```diff
@@ -45,20 +46,32 @@ class Channel {
...
+    const mergedHeaders = new Headers(headers)
+    if (this.headers && typeof this.headers.entries === 'function') {
+      for (const [key, value] of this.headers.entries()) {
+        // Only add headers from this.headers that don't exist in the request headers
+        if (!(key in mergedHeaders)) {
+          mergedHeaders.set(key, value)
+        }
+      }
+    }
+
```
alanshaw, line 72:
```suggestion
    headers = { ...this.headers, ...headers }
```

alanshaw (changes requested):

packages/transport/src/http.js
fforbeck: That's how you recognize a real PRO!

alanshaw (approved):

PR #378, opened 2025-05-12
fix!: did bytes schema

alanshaw:
This PR fixes the `didBytes` schema so that it returns the type it is designed to read - `Uint8Array`. Concretely this changes the return type from `read` to be a `PrincipalView`, which is a subclass of `Uint8Array`.

The issue is that while this schema (in its current form) can read and validate data encoded as a `Uint8Array`, you cannot _create_ an invocation/delegation with a value of a `Uint8Array` - it expects a string DID:

```
Type 'Uint8Array<ArrayBufferLike>' is not assignable to type '`did:${string}:${string}` & `did:${string}` & Phantom<{ protocol: "did:"; }>'.
```

If you pass a string, it will be encoded as a string, which is not the value that can be read by the same schema.

In real terms, I'm trying to create a location claim with a byte encoded space DID but I cannot 😢

In hindsight, this schema type should perhaps be called `principal`. I'd be open to changing this PR to add that and deprecate `didBytes` instead if folks feel strongly.

### DIFF
--- a/packages/client/test/services/util.js
+++ b/packages/client/test/services/util.js
@@ -1,6 +1,6 @@
 export const ok =
   /** @type {<T, Args extends []|[T]>(...args:Args) => Args extends [T] ? {ok:true, value:T extends undefined ? null : T} : {ok:true, value:null}}} */ (
-    (value) => (value == undefined ? Ok : { ok: true, value })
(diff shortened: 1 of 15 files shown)


hannahhoward (approved):
Lgtm and I prefer the rename to Principal if we have bandwidth to do it.

PR #379, opened 2025-05-13
feat: add principal schema

alanshaw:
This PR adds `Schema.principal()` as a replacement to `Schema.didBytes()`.

The `Schema.didBytes()` schema has issues as described here https://github.com/storacha/ucanto/pull/378. This change allows those issues to be fixed without a breaking change. This PR also deprecates `Schema.didBytes()` via jsdoc comments.


alanshaw:
Merging this as there's no code changes from https://github.com/storacha/ucanto/pull/378, which is already approved.

PR #381, opened 2025-05-29
feat!: resolve multiple verifiers

fforbeck:
### Support Multiple Verifiers in DID Resolution

Added support for multiple verifiers in the DID Key resolution process. Previously, the `config.resolveDIDKey` method would resolve a single verifier, which is fine. However, if we decide to resolve `did:key` from a `did:plc` document, we need to resolve all verifiers traversed the returned list to find one that successfully verifies the signature.

## Main Changes
- Modified the `verifyAuthorization` function in `packages/validator/src/lib.js` to handle multiple verifiers returned by `resolveDIDKey`
- Added support for iterating through multiple verifiers and attempting verification with each one
- Added new test cases to verify the behavior with multiple verifiers

Related: https://github.com/storacha/project-tracking/issues/463

### DIFF
--- a/packages/interface/src/lib.ts
+++ b/packages/interface/src/lib.ts
@@ -1092,8 +1092,8 @@ export interface PrincipalParser {
  */
 export interface PrincipalResolver {
   resolveDIDKey?: (
-    did: UCAN.DID
-  ) => Await<Result<DIDKey, DIDKeyResolutionError>>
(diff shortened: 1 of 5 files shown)


alanshaw (changes requested):
Needs to be marked as a breaking change for interface and validator.

PR #386, opened 2025-07-25
docs: rectify validator README

Nkovaturient:
## Problem:

-  The validator [README](https://github.com/storacha/ucanto/blob/main/packages/validator/README.md#example-usage) has a example usage showcasing non-existent `validate` function, hence rectifying it with correct methods:

## Solution:

- Define capabilities using `capability` from @ucanto/validator.
- Invoke using the `capability's .invoke()` method.
- Validate using `access` from @ucanto/validator.

### DIFF
--- a/packages/validator/README.md
+++ b/packages/validator/README.md
@@ -22,13 +22,12 @@ npm install @ucanto/validator
 
 ## Example Usage
 ```ts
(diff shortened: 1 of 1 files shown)


alanshaw (changes requested):

alanshaw (approved):
LGTM

Issue #387, opened 2025-08-15
docs: verify and update README examples for latest releases


Dhruv-Varshney-developer:
## What?
A user reached out with the following query on [discord](https://discord.com/channels/1247475892435816553/1405283951601389619):
"Hello Support Team. Is there any updated `Read me` files? All Read Me data were wrong for latest version in the https://github.com/storacha/ucanto `@ucanto/transport @ucanto/server`, etc. I think `import * as CBOR from '@ucanto/transport/cbor`'; is no more exist in the latest version at v9.2.0 And the functions `Server.create(), Client.connect() `should be replaced with new info."

### Tasks to be completed 
- [ ] Test all examples and code snippets in READMEs.
- [ ] Check if they are updated as per latest version. 
- [ ] If not, update them to reflect current API.
- [ ] Add any missing examples and improve the documentation.

NiKrause:
i made a new PR which should include this https://github.com/storacha/ucanto/pull/409

PR #388, opened 2025-08-21
docs: Fix outdated READMEs for @ucanto/client and @ucanto/transport

Dhruv-Varshney-developer:
Partially completes work for #387 - Addresses user report of broken imports and wrong function calls in documentation.

### Changes:

1. **@ucanto/transport:**
The old README had imports that don't exist (`@ucanto/transport/cbor`) and showed wrong function calls. Rewrote it with actual working code that matches what's in the package. Also updated the descriptions to accurately reflect what the package actually does instead of generic text. Now developers can copy-paste and run examples instead of getting import errors.

2. **@ucanto/client:**
The old README used outdated API calls (`ed25519.Verifier.parse()`) and had incomplete examples with undefined variables. Fixed all the function calls to match current API and added the missing connection setup code. Now the examples actually work when you run them.

#### Both packages:

- Added test keys from `packages/client/test/fixtures.js` so examples work immediately
- Examples snippets can be run(tested) directly from https://github.com/Dhruv-Varshney-developer/ucanto-storacha
- Reflects latest exports and method signatures.

Before: Developers tried the README examples and got errors because the code was wrong.
After: Examples run successfully and show how to actually use these packages.

### DIFF
--- a/packages/client/README.md
+++ b/packages/client/README.md
@@ -78,24 +78,58 @@ const [receipt] = await connection.execute(invocation)
 console.log(receipt.out.error ? 'Failed:' : 'Success:', receipt.out)
 ```
 
-### Run it:
+## Setup Instructions
 
-⚠️ These are test keys. Replace them with your own service ID and client keypair before using in production.
+### Environment Variables
+
+**AGENT_PRIVATE_KEY**
+Set the key your client should use to sign UCAN invocations. You can generate Ed25519 keys with the ucanto library.
+
+#### Usage
+
+Create a file called `generate-keys.js`:
+
+```javascript
+import { ed25519 } from '@ucanto/principal'
+
+async function generateKeys() {
+  const keypair = await ed25519.generate()
+  
+  const privateKey = ed25519.format(keypair)
+  
+  console.log('AGENT_PRIVATE_KEY=' + privateKey)
+}
+
+generateKeys().catch(console.error)
+```
+
+Then run it:
 
 ```bash
-SERVICE_ID="MgCYKXoHVy7Vk4/QjcEGi+MCqjntUiasxXJ8uJKY0qh11e+0Bs8WsdqGK7xothgrDzzWD0ME7ynPjz2okXDh8537lId8=" \
-AGENT_PRIVATE_KEY="MgCZT5vOnYZoVAeyjnzuJIVY9J4LNtJ+f8Js0cTPuKUpFne0BVEDJjEu6quFIU8yp91/TY/+MYK8GvlKoTDnqOCovCVM=" \
-node example.js
+node generate-keys.js
 ```
 
+**SERVICE_DID**
+Set the DID of the service you want to connect to. Check the service's documentation for their public DID.
 
-### Connecting to a Real Service
+**SERVICE_URL** (Optional)
+If you're connecting to a custom service, set both `SERVICE_DID` and `SERVICE_URL` environment variables.
 
-Replace the `mockFetch` in the example with the real `fetch` and a valid service URL:
 
-```js
-channel: HTTP.open({ url: new URL('https://api.example.com'), fetch })
+For example, Storacha has following `SERVICE_DID` and `SERVICE_URL`:
+
+```bash
+# Storacha uses these default values:
+SERVICE_DID="did:web:up.storacha.network"
+SERVICE_URL="https://up.storacha.network"
 ```
+
+Set your environment variables like so:
+```bash
+AGENT_PRIVATE_KEY="your_generated_private_key_here" \
+SERVICE_DID="did:key:service_provider_did_here" \
+```
+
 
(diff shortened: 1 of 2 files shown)


packages/client/README.md
```diff
@@ -22,28 +23,80 @@ npm install @ucanto/client
...
+import { Receipt, Message } from '@ucanto/core'
 
-const service = ed25519.Verifier.parse(process.env.SERVICE_ID);
-const issuer = ed25519.Signer.parse(process.env.CLIENT_KEYPAIR);
+const service = ed25519.parse(process.env.SERVICE_ID)
+const issuer = ed25519.parse(process.env.CLIENT_KEYPAIR)
 
-const invocation = await Client.invoke({
+// Mock fetch that simulates a UCAN service
+const mockFetch = async (url, input) => {
```
alanshaw, line 37: I'm not sure if it's a good idea for a mock to be in the README, perhaps in a runnable code _example_ in the repo. IMHO the README should shown how to use the library against an already running server - i.e. just the code you'd need to use the client. I'd link to the _server_ README for actually setting up and running a server component.

packages/client/README.md
```diff
@@ -22,28 +23,80 @@ npm install @ucanto/client
...
+```js
+import * as Client from '@ucanto/client'
+import * as HTTP from '@ucanto/transport/http'
+import { CAR } from '@ucanto/transport'
+import { ed25519 } from '@ucanto/principal'
+import { Receipt, Message } from '@ucanto/core'
 
-const service = ed25519.Verifier.parse(process.env.SERVICE_ID);
-const issuer = ed25519.Signer.parse(process.env.CLIENT_KEYPAIR);
+const service = ed25519.parse(process.env.SERVICE_ID)
```
alanshaw, line 33: A client will typically not have the private key of the service, we need to tech them how to parse a DID (the public key) here.

packages/client/README.md
```diff
@@ -22,28 +23,80 @@ npm install @ucanto/client
...
+const issuer = ed25519.parse(process.env.CLIENT_KEYPAIR)
```
alanshaw, line 34: I'd probably not call this "keypair" - it's the private key, from which a public key may be derived.

Suggestions:
* `PRIVATE_KEY`
* `AGENT_PRIVATE_KEY`

packages/client/README.md
```diff
@@ -22,28 +23,80 @@ npm install @ucanto/client
...
+console.log(receipt.out.error ? 'Failed:' : 'Success:', receipt.out)
+```
+
+### Run it:
+
+⚠️ These are test keys. Replace them with your own service ID and client keypair before using in production.
+
+```bash
+SERVICE_ID="MgCYKXoHVy7Vk4/QjcEGi+MCqjntUiasxXJ8uJKY0qh11e+0Bs8WsdqGK7xothgrDzzWD0ME7ynPjz2okXDh8537lId8=" \
+CLIENT_KEYPAIR="MgCZT5vOnYZoVAeyjnzuJIVY9J4LNtJ+f8Js0cTPuKUpFne0BVEDJjEu6quFIU8yp91/TY/+MYK8GvlKoTDnqOCovCVM=" \
```
alanshaw, line 87: I would not provide these, but instruct users on how they can generate their own. It prevents them from getting accidentally used in production services.

packages/transport/README.md
```diff
@@ -21,12 +22,98 @@ npm install @ucanto/transport
...
+import { ed25519 } from '@ucanto/principal'
+import { invoke, Message, Receipt } from '@ucanto/core'
 
-const encoded = CAR.encode({ invocations: [] });
-const decoded = CBOR.decode(encoded);
+const service = ed25519.parse(process.env.SERVICE_ID)
+const issuer = ed25519.parse(process.env.CLIENT_KEYPAIR)
+
+// Mock fetch that simulates a UCAN service
+const mockFetch = async (url, init) => {
```
alanshaw, line 35: Again, I'd save this for a runnable example in the repo, or better yet a runnable example with an actual server. Mocking the server like this is not great even for tests as it bypasses all delegation chain validation.

Note: in tests, you don't even need a HTTP server running - a Ucanto server _is_ a channel, so you can pass it to the client as the channel to use.

packages/client/README.md
Dhruv-Varshney-developer: [2a89648](https://github.com/storacha/ucanto/pull/388/commits/2a896485b2ab2a9d202a27e1441b2efd539917a6)

packages/client/README.md
Dhruv-Varshney-developer: I have a question here: how can a developer get SERVICE_DID if we don't provide neither the private key nor the DID?

packages/client/README.md
Dhruv-Varshney-developer: https://github.com/storacha/ucanto/pull/388/commits/5dc05d33f190086481fd4a63f850e12e4100c446

NiKrause:
i think this PR can be closed since of merged: https://github.com/storacha/ucanto/pull/407 (which includes that PR)

PR #408, opened 2026-02-11
feat(principal)!: webcrypto ed25519 signer support for non-extractable agent keys

NiKrause:
## Summary
- switch `@ucanto/principal` Ed25519 signer implementation to WebCrypto
- default Ed25519 `generate()` to non-extractable keys while preserving extractable mode when explicitly requested
- remove direct `@noble/ed25519` signing path from Ed25519 signer
- add PKCS8/JWK conversion logic for Ed25519 derive/generate extractable flows
- add coverage-focused tests for extractable/non-extractable behavior and error paths
- update server receipt assertion to compare issuer by DID to avoid brittle object-shape comparison

## Context
This aligns with storacha/upload-service#658, which requires browser agent key generation to move to non-extractable WebCrypto Ed25519 keys.

## Breaking Changes
⚠️ **Libs and tests which required extractable keys now have to explicitly request it with `generate({ extractable: true })`**

Previously, keys were extractable by default. Now:
- **Default behavior**: `generate()` produces non-extractable keys
- **Extractable keys**: Must explicitly use `generate({ extractable: true })`
- **Impact**: Code that relied on `.secret()` or `.encode()` access will need updates

## Remark
Many changes are just prettier formatting changes

---

Co-Authored-By: NiKrause

### DIFF
--- a/packages/principal/package.json
+++ b/packages/principal/package.json
@@ -29,7 +29,6 @@
   "dependencies": {
     "@ipld/dag-ucan": "^3.4.5",
     "@noble/curves": "^1.2.0",
-    "@noble/ed25519": "^1.7.3",
     "@noble/hashes": "^1.3.2",
     "@ucanto/interface": "workspace:^",
     "multiformats": "^13.3.1",
--- a/packages/principal/src/ed25519/signer.js
+++ b/packages/principal/src/ed25519/signer.js
@@ -1,9 +1,8 @@
-import * as ED25519 from '@noble/ed25519'
 import { webcrypto } from 'one-webcrypto'
 import { varint } from 'multiformats'
 import * as API from './type.js'
 import * as Verifier from './verifier.js'
-import { base64pad } from 'multiformats/bases/base64'
+import { base64pad, base64url } from 'multiformats/bases/base64'
 import * as Signature from '@ipld/dag-ucan/signature'
 import * as Signer from '../signer.js'
 export * from './type.js'
@@ -19,6 +18,11 @@ const PRIVATE_TAG_SIZE = varint.encodingLength(code)
 const PUBLIC_TAG_SIZE = varint.encodingLength(Verifier.code)
 const KEY_SIZE = 32
 const SIZE = PRIVATE_TAG_SIZE + KEY_SIZE + PUBLIC_TAG_SIZE + KEY_SIZE
+const ALG = { name: 'Ed25519' }
+const PKCS8_PREFIX = Uint8Array.from([
+  0x30, 0x2e, 0x02, 0x01, 0x00, 0x30, 0x05, 0x06, 0x03, 0x2b, 0x65, 0x70, 0x04,
+  0x22, 0x04, 0x20,
+])
 
 export const PUB_KEY_OFFSET = PRIVATE_TAG_SIZE + KEY_SIZE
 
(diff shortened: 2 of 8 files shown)


hannahhoward (approved):
Wow. So this is pretty great. Also great ed25519 FINALLY landed in webcrypto.... I followed but it was taking so long I lost track several times :P I think I'm good to approve it as is.

Just to confirm, re: extractable / non-extractable -- this isn't ultimately a breaking change cause the API.Signer returned by generate didn't provider access to the secret() method so you'd need a type cast for that anyway. 

Also, does it make sense to rebuild the old ED25519Signer on an actual webcrypto key instead of importing on sign? Is there a perf cost to importing every time we need to sign? (I assuming not a significant one cause the compute cost comes from actually signing as opposed to importing)

Anyway, would love some color on the answers to these questions, but I am otherwise good to merge, and thank you so much for this contribution.

NiKrause:
Thank you @hannahhoward for looking into this! 

First of all, I enabled type checks on all branches ``**`` and added the same typescript fixes from
https://github.com/storacha/ucanto/pull/410 to this PR too otherwise they don't pass on the new failing main branch. 

Otherwise I answer inline: 
> Just to confirm, re: extractable / non-extractable -- this isn't ultimately a breaking change cause the API.Signer returned by generate didn't provider access to the secret() method so you'd need a type cast for that anyway.

There's a good example here in the upload-service, which accessed and again accesses .secret directly from signer (ed25519.secret) see: https://github.com/storacha/upload-service/pull/667/changes#diff-e0788dca19d1076e1e2a937c09bf86711f6d4763400b47a743d88849b16abbbf
Also the previous Signer had a secret method: https://github.com/storacha/ucanto/blob/main/packages/principal/src/ed25519/signer.js#L155

Libs and tests which required extractable keys now have to enforce it with ```generate({ extractable: true })```  the upload-service. That is probably a breaking change...

> Also, does it make sense to rebuild the old ED25519Signer on an actual webcrypto key instead of importing on sign? Is there a perf cost to importing every time we need to sign? (I assuming not a significant one cause the compute cost comes from actually signing as opposed to importing)

We have now the "old" ED25519Signer and a new UnextractableEd25519Signer mainly for compatibility reasons. 
If older code would still request .secret or .encode without enforcing it now with ```generate({ extractable: true })```, it throws because private key bytes cannot be exported. Not so much because of performance improvements.

We could remove the ED25519Signer completely but e.g. the toArchive() function is needed elsewhere in other imlementing projects signing UCANs.

If we would use only one ED25519Signer, it would pretend it has access to .secret() and .encode(). Would be possible but seems worse design as of my research. 

edited 2026-02-18 13:58 CET

NiKrause:
I force-pushed a slimmed version of this PR (now only the WebCrypto Ed25519 changes + minimal related tests/lockfile). Please merge the two prerequisite PRs first, since they contain CI/Playwright fixes and formatting-only cleanup:
- #411 (CI/Playwright Windows + runner adjustments)
- #412 (prettier-only formatting)

Once those are merged, this PR should be clean and focused.

NiKrause:
Update: I added a Node 20+ matrix change to the CI/Playwright prerequisite PR (#411). This is required for WebCrypto Ed25519 () to work reliably in CI. Once #411 and #412 merge, this PR should be green without further changes here.

NiKrause:
Update: I added a Node 20+ matrix change to the CI/Playwright prerequisite PR (#411). This is required for WebCrypto Ed25519 (subtle.generateKey({ name: 'Ed25519' })) to work reliably in CI. Once #411 and #412 merge, this PR should be green without further changes here.

PR #407, opened 2025-11-18
docs: combined 388 392

NiKrause:
**Addresses @alanshaw recommendations from PR #388**

This branch combines PRs #388 and #392 and addresses all review comments from @alanshaw.

### Changes

**README updates:**
- Removed mock fetch functions from `@ucanto/client` and `@ucanto/transport` READMEs
- Updated examples to show real server usage (via environment variables)
- Changed `ed25519.Verifier.parse()` to `DID.parse()` for service DIDs (as suggested)
- Added notes about using server-as-channel for testing, with links to server README

**Test improvements:**
- Converted `packages/client/test/client.spec.js` from mock fetch to server-as-channel pattern
- Tests now validate delegation chains (no bypass)
- Added `@ucanto/server` and `@ucanto/validator` to client devDependencies

### Benefits

- Examples work with real servers (no mocks)
- Tests properly validate UCAN delegation chains
- Follows recommended patterns from the server package
- All tests passing with 100% coverage maintained

### Related

- Addresses review comments from: https://github.com/storacha/ucanto/pull/388
- Should complete work for: #387 (which was respected in #392)

PR #410, opened 2026-02-18
test: move server-channel integration to server and fix client typechecks

NiKrause:
## Summary
- move server-as-channel integration tests from `packages/client/test/client.spec.js` to `packages/server/test/client-channel.spec.js`
- keep client tests package-local (no `@ucanto/server` / `@ucanto/validator` imports)
- add local client execute/decode tests to restore `packages/client` 100% coverage gates
- run package workflows on pushes to all branches (`push.branches: ["**"]`) for `client`, `core`, `interface`, `principal`, `server`, `transport`, and `validator`

## Why
The previous client test changes mixed server integration concerns into the client package, which caused fragile TypeScript behavior in isolated package checks.

This PR keeps boundaries clear:
- server package owns server-channel integration behavior
- client package keeps self-contained tests and coverage

## Why checks failed after merge previously
On PR #407, package CI typecheck jobs did not run on the PR head commit (only `Validate PR title` was recorded). On merge commit `60c70da`, push workflows ran package typechecks and surfaced failures.

So the issue was check coverage/enforcement at PR time (branch protection/ruleset/settings), not new source changes introduced between PR creation and merge.

## Remark
This PR now also removes `main`-only `pull_request.branches` constraints for all package workflows
---


.github/workflows/client.yml
```diff
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
   push:
     branches:
-      - main
+      - '**'
```
alanshaw: I think this is perhaps not quite the right change. I think the `pull_request.branches` config needs to be removed (or changed to `'**'`) instead. This will run these workflows on any push to any branch, whereas I think we only really need it to run when opening (or amending) a PR that we want merged. This is consistent with other repos we have in the organization.

packages/server/test/server-channel.spec.js
alanshaw, line 1: I'd perhaps name this `server-channel.js` since the server is acting as the channel in the tests.

alanshaw (changes requested):

packages/server/test/server-channel.spec.js
NiKrause: i renamed it.

.github/workflows/client.yml
NiKrause: ok. that is naturally better. I re-removed it from pull_request.branches and now only main branch o push is triggered.

alanshaw (approved):

